PR #4, opened 2024-04-08
Fix rollback issue

joshklop:
The version of goleveldb that we were using before returned `nil` instead of `[]byte{}`. Unfortunately, this changes the behavior in v0.20.0 of `github.com/cosmos/iavl`. The sdk and iavl interpret `nil` as "doesn't exist", which meant that the iavl mistakenly couldn't rollback to past states if the tree was empty (which happened for modules that had an empty genesis state and no state updates thereafter).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -175,6 +175,13 @@ replace (
 	github.com/ethereum-optimism/optimism v1.4.2 => github.com/polymerdao/optimism-dev v1.4.2-polymer-1.1
 	// use OP's fork of op-geth
 	github.com/ethereum/go-ethereum v1.13.5 => github.com/ethereum-optimism/op-geth v1.101305.2-rc.2.0.20240117002010-d5f142e54a0a
+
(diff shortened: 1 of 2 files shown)


NiloCK (approved):
Devil's advocacy: pinning a specific version as _non buggy_ isn't very future proof.

Do we know if, eg, dependabot is configured here to flag issues that arise?

joshklop:
> Devil's advocacy: pinning a specific version as non buggy isn't very future proof.

Completely agree. The package itself appears to be abandoned. I really hope we can move to a different db at some point, as a long-term solution. The main thing precluding us from doing that right now is the tx indexing, which depends on a comet-style db backend (to fix that, we'll need to vendor and rewrite [this function](https://github.com/cometbft/cometbft/blob/c3acea2436f4fc0d2cbbb8b27ada97ab3706892f/internal/state/txindex/kv/kv.go#L419-L536)) :(

So we're stuck with this until we prioritize a db refactor.

PR #11, opened 2024-04-11
Remove the cli in favor of a proper `node` package

joshklop:
## Summary by CodeRabbit

- **New Features**
  - Replaced build target in Makefile with new test target for running tests.
  - Introduced HTTP server management package with start and stop functionality.
- **Refactor**
  - Refactored main package into `node` package for improved node initialization and operation.
  - Updated packages and structs to enhance code structure and resolve dependencies.
- **Tests**
  - Added new tests for blockchain node operations and adjusted existing tests for better coverage.
- **Chores**
  - Updated Makefile to handle formatting tools and removed specific build targets.
  - Modified GitHub Actions workflow for linting with a custom script for consistency.
- **Documentation**
  - Updated README.md to reflect changes in local component execution and test commands.


node/http.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (h *httpService) Start() error {
+	go func() {
+		if err := h.srv.Serve(h.listener); !errors.Is(err, http.ErrServerClosed) {
+			h.errCh <- fmt.Errorf("serve http: %v", err)
+		}
+	}()
+	return nil
+}
```
NiloCK, line 25: No error is actually returned here, but instead is piped into `errCh`. Could simplify calling code via:

```suggestion
func (h *httpService) Start() {
	go func() {
		if err := h.srv.Serve(h.listener); !errors.Is(err, http.ErrServerClosed) {
			h.errCh <- fmt.Errorf("serve http: %v", err)
		}
	}()
}
```

node/node.go
```diff
@@ -93,50 +65,73 @@ func run(ctx context.Context) (err error) {
...
+	<-ctx.Done()
 	return nil
 }
 
+func runAndWrapOnError(existingErr error, msg string, fn func() error) error {
+	if runErr := fn(); runErr != nil {
+		if existingErr == nil {
+			return runErr
+		}
+		runErr = fmt.Errorf("%s: %v", msg, runErr)
```
NiloCK, line 129: I think this probably wants to be:

```suggestion
		runErr = fmt.Errorf("%s: %v", msg, runErr)
		if existingErr == nil {
			return runErr
		}
```

As is, if the pass function produces an error with no passed, bundled, error, the context string that describes what's being attempted gets lost.

NiloCK (approved):
This LGTM and it's nice to have a basis for automated tests.

Presumed next steps are to expand node_test with things hitting the EngineAPI service.

docs nit: Might be nice to scrub the `Development` section of the readme to just "run `make test`". Good to keep in sync.

edit: I see the readme edit is made in #12.

joshklop:
> docs nit: Might be nice to scrub the Development section of the readme to just "run make test". Good to keep in sync. edit: I see the readme edit is made in #12.

You're right, the change should be in this PR. Should be fixed now. Thanks!

node/node.go
joshklop: Great catch, thanks!

node/http.go
joshklop: Nice. I modified the http service to use `Run(ctx context.Context) error` instead. I think it's clearer and it pushes concurrency up to the next level of abstraction.

PR #12, opened 2024-04-12
Add e2e test

joshklop:
Depends on #11 

## Summary by CodeRabbit

- **New Features**
  - Added new end-to-end testing capabilities and tools for Ethereum node interaction.
  - Introduced utility functions for enhanced error handling and context manipulation.
  - Updated development prerequisites and instructions in the README.
- **Documentation**
  - Updated README to reflect new development prerequisites and testing instructions.
- **Tests**
  - Expanded test coverage with new end-to-end tests for Ethereum node components and utility functions.
- **Chores**
  - Updated workflow configurations to include new end-to-end testing jobs.
- **Refactor**
  - Enhanced error handling in node operations by integrating new utility functions.


e2e/op_stack.go
```diff
@@ -0,0 +1,282 @@
...
+	LogWithPrefix(prefix string, r *log.Record)
+	AfterStartup()
+	BeforeShutdown()
+}
+
+type OPStack struct {
+	// The http address (without the http://) of the Ethereum EL node.
+	l1Address string
+	// The websocket address (without the ws://) of the execution engine.
+	engineAddress string
```
NiloCK, line 47: maybe, for consistency w/ other listings of these addresses:
```suggestion
    // the websocket address (without the we://) of the Monomer execution engine.
	monomerAddress string
```

e2e/stack.go
```diff
@@ -0,0 +1,213 @@
...
+		cancel(fmt.Errorf("run cmd: %v", s.runCmd(
+			ctx,
+			"anvil",
+			"--port", l1Port,
+			"--order", "fifo",
+			"--disable-block-gas-limit",
+			"--gas-price", "0",
+			"--block-time", fmt.Sprint(s.l1BlockTime.Seconds()),
+		)))
+	})
```
NiloCK, line 80: The nesting function calls & implicit hanging function made this a bit hard to parse.

Intent is probably clearer with something like this: 

```suggestion
    wg.Go(func() {
		anvilErr := s.runCmd(
			ctx,
			"anvil",
			"--port", l1Port,
			"--order", "fifo",
			"--disable-block-gas-limit",
			"--gas-price", "0",
			"--block-time", fmt.Sprint(s.l1BlockTime.Seconds()),
		)
		if anvilErr != nil {
			cancel(fmt.Errorf("run cmd: %v", anvilErr))
		}
	})
```

e2e/stack.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		return fmt.Errorf("dial anvil: %v", err)
+	}
+	anvil := NewAnvil(anvilRPCClient)
+	privKey, err := crypto.GenerateKey()
+	if err != nil {
+		return fmt.Errorf("generate key: %v", err)
+	}
+	address := crypto.PubkeyToAddress(privKey.PublicKey)
+	if err := anvil.SetBalance(ctx, address, 1000000000000000000); err != nil { //nolint:gomnd
```
NiloCK, line 97: Better still to have something like `1ETH` run through a converter, but I couldn't immediately find those utilities in geth.

```suggestion
	if err := anvil.SetBalance(ctx, address, 1_000_000_000_000_000_000); err != nil { //nolint:gomnd
```

e2e/stack.go
```diff
@@ -0,0 +1,213 @@
...
+		return fmt.Errorf("set balance: %v", err)
+	}
+
+	// Deploy L1 contracts.
```
NiloCK, line 101:
```suggestion
	// Deploy OP contracts to L1.
```

e2e/stack.go
```diff
@@ -0,0 +1,213 @@
...
+	latestL1Block, err := anvil.BlockByNumber(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("get the latest l1 block: %v", err)
+	}
+
+	// Run Monomer.
+	const l2ChainID = 901
+	wg.Go(func() {
+		cancel(fmt.Errorf("run monomer: %v", runMonomer(ctx, latestL1Block.Time(), l2ChainID, s.monomerAddress)))
+	})
```
NiloCK, line 124: as above:

```suggestion
		wg.Go(func() {
		monomerErr := runMonomer(ctx, latestL1Block.Time(), l2ChainID, s.monomerAddress)
		if monomerErr != nil {
			cancel(fmt.Errorf("run monomer: %v", monomerErr))
		}
	})
```

e2e/stack.go
```diff
@@ -0,0 +1,213 @@
...
+		return fmt.Errorf("dial monomer: %v", err)
+	}
+	monomerClient := NewMonomer(monomerRPCClient)
+	l2GenesisBlockHash, err := monomerClient.GenesisHash(ctx)
+	if err != nil {
+		return fmt.Errorf("get Monomer genesis block hash: %v", err)
+	}
+
+	// Get deploy config and rollup config.
+	const networkName = "hardhat"
```
NiloCK, line 139: ?

```suggestion
	const networkName = "anvil"
```

e2e/op_stack.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+func (op *OPStack) Run(parentCtx context.Context) (err error) {
+	var wg conc.WaitGroup
+	defer wg.Wait()
+	ctx, cancel := context.WithCancelCause(parentCtx)
+	defer func() {
+		op.eventListener.BeforeShutdown()
+		cancel(err)
+		err = utils.Cause(ctx)
```
NiloCK, line 85:
```suggestion
		err = utils.Cause(ctx)
		cancel(err)
```

e2e/op_stack.go
```diff
@@ -0,0 +1,282 @@
...
+		cancel(op.runProposer(ctx, anvil, txManagerConfig))
+	})
+
+	wg.Go(func() {
+		cancel(op.runBatcher(ctx, anvil, txManagerConfig))
+	})
+
+	op.eventListener.AfterStartup()
+	<-ctx.Done()
+	return nil
```
NiloCK, line 130: Should these be `return err` for most of these functions? They all declare named returns, and some of them set the `err` variable during a deferred cleanup.

e2e/stack.go
```diff
@@ -0,0 +1,213 @@
...
+			"--gas-price", "0",
+			"--block-time", fmt.Sprint(s.l1BlockTime.Seconds()),
+		)))
+	})
+	if !utils.AddressIsReady(s.l1Address, numConnectAttempts) {
+		return fmt.Errorf("could not connect to anvil at address %s", s.l1Address)
+	}
+
+	// Fund an account.
+	anvilHTTPAddress := "http://" + s.l1Address
```
NiloCK, line 86: There are a bunch of instances of this rpc client / dialContext through `stack` and `op_stack`.

Should they instead live inside of, eg, `NewAnvil(address string)` which could return an error? Might make the "main execution paths" a bit less noisy and easier to read. Similar for other client constructors.

NiloCK (approved):
This is great. Left a bunch of items. Some small and quick, others maybe will take a little thought.

e2e/stack_test.go
```diff
@@ -0,0 +1,136 @@
...
+	})
+
+	var wg conc.WaitGroup
+	defer wg.Wait()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	wg.Go(func() {
+		require.NoError(t, stack.Run(ctx))
+	})
+	require.True(t, utils.AddressIsReady(monomerAddress, 10000000))
```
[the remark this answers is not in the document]
joshklop: Haha, yeah I should probably add a comment for this one. It's deliberate. What do you think about the following?

> To avoid flaky tests, hang until the Monomer server is ready. We rely on the `go test` timeout to ensure the tests don't hang forever (default is 10 minutes).

utils/utils.go
```diff
@@ -0,0 +1,44 @@
...
+	for range numConnectAttempts {
+		conn, err := net.Dial("tcp", address)
+		if err == nil {
+			conn.Close()
+			return true
+		}
+		time.Sleep(250 * time.Millisecond) //nolint:gomnd
+	}
+	return false
+}
```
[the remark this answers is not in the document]
joshklop: Love this!

e2e/stack_test.go
```diff
@@ -0,0 +1,136 @@
...
+		block, err := monomerClient.BlockByNumber(ctx, new(big.Int).SetUint64(i))
+		require.NoError(t, err)
+		txs := block.Transactions()
+		require.Len(t, txs, 1)
+		if tx := txs[0]; !tx.IsDepositTx() {
+			txBytes, err := tx.MarshalJSON()
+			require.NoError(t, err)
+			require.Fail(t, fmt.Sprintf("expected tx to be deposit tx: %s", txBytes))
+		}
+	}
```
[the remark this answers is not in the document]
joshklop: This is intentionally placed after the test that ensures monomer is synced to at least `targetHeight`.

e2e/stack.go
joshklop: Yeah that's a good idea. I also looked for this in geth but couldn't find it :)
I'll think about adding a converter func to make things clearer.

e2e/stack.go
joshklop: I'll have to add a comment for this; it's deliberate. I forget the details at this moment, but anvil is meant to be a drop-in replacement for hardhat in the op scripts IIRC.

e2e/op_stack.go
joshklop: Not sure I follow this change. If `err != nil` and the context hasn't been canceled, won't this cause us to incorrectly return `nil`?

e2e/op_stack.go
joshklop: `err` will be set to the value that the function returns with. I find `return nil` to be more explicit than `return err`, since the latter makes one think that `err != nil` somehow. Your thoughts?

e2e/stack.go
NiloCK: :+1: 

Thought there was a chance this was true, but I recalled you had swapped backends a couple of times during development and figured it might have been a leftover.

e2e/stack.go
joshklop: Good point. I don't like the address handling either. I'll think about how to improve it.

I didn't make `NewAnvil` take the `address` since that would require it to create an rpc client internally, violating dependency injection. That might be a bit pedantic on my part though.

utils/utils.go
```diff
@@ -0,0 +1,44 @@
...
+		runErr = fmt.Errorf("%s: %v", msg, runErr)
+		if existingErr == nil {
+			return runErr
+		}
+		return fmt.Errorf("operation failed: %v, previous error: %w", runErr, existingErr)
+	}
+	return existingErr
+}
+
+func Cause(ctx context.Context) error {
```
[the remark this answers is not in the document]
joshklop: I love the comment. Will add.

I personally prefer the `Cause` name since `utils.Cause` vs. `context.Cause` should be enough to signal that something is different, while still being closely related.

e2e/op_stack.go
NiloCK: I wrote that believing that the below would return `nil`. It doesn't, which honestly I find a little disturbing, but I'm adjusting.

I think it can stay as is, or possibly the return statement could be dropped altogether.

```
func impErr() (err error) {
	defer func() {
		err = errors.New("this is an error")
	}()

	return nil
}
```

joshklop: I see. The rabbit ai bot thought it was confusing as well.

I find it tough to return coherent error messages when dealing with concurrency in general. Using named returns, defers, and a `context.CancelCauseFunc` is the best solution I could think of.

utils/utils.go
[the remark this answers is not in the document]
NiloCK: I think that's OK, but, devil's advocacy: it's a bit of a tease to signal that something is different, while witholding potential clue about _what_ is different. It's in the comment, but best to surface as much digestible info as early as possible.

I didn't really love `External`, but no other adjectives sprang to mind. `Injected`? `Manual`? `Imperative`?

utils/utils.go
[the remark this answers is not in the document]
joshklop: I may make some more modifications. I now realize there should be a `context` here, for example.

e2e/op_stack.go
joshklop: Looking at this again, I think "execution engine" is sufficient. The OP CL is agnostic to the engine implementation.

utils/utils.go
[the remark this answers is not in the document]
joshklop: Really appreciate the push-back here.

> it's a bit of a tease to signal that something is different, while witholding potential clue about what is different. 

100% agree. However, in this case I believe the reason is subtle enough that any adjective risks obscuring the meaning even more. I'll move forward with the comment for now, but if you feel strongly, I'll change the name (another candidate is `UnderlyingCause`).

e2e/stack.go
joshklop: Fixed this by wrapping `url.URL`.

e2e/stack.go
joshklop: Nice catch! There's also a bug in my original version where we'll end up wrapping a `nil` error. Should be fixed in the latest commit. That proves it was indeed too confusing.

joshklop:
@NiloCK Thanks for the insightful review. I attempted to address the comments in the latest commit. I plan to rebase and force-push prior to merging, but I figured that because this is a hefty PR, a new commit is easier to review.

The most interesting changes are probably in the new `e2e/url` package, which should simplify address handling.

PR #8, opened 2024-04-09
Payload store refactors

NiloCK:
PR:

1. Includes a few small refactors to the payloadStore.
    - removes its mutex (safety provided further up the chain)
    - removes its interface (interface abstraction is overkill for now)
    - removes some redundant logic
2. Adds some functionality to the payloadStore
    - `Remove`, `Clear`
3. Fixes existing functionality
    - rollback now rolls back rather than clears the store.
4. Adds tests for the above.

## Summary by CodeRabbit

- **New Features**
	- Enhanced payload management capabilities including the ability to remove, rollback, and clear payloads.
  
- **Bug Fixes**
	- Improved logic in payload addition and current payload retrieval functions for better accuracy and performance.

- **Refactor**
	- Updated payload store interface and method signatures for streamlined operations.

- **Tests**
	- Added and modified tests to ensure robust handling of payloads, including testing new functionalities like clearing and batch removal based on specific criteria.

PR #16, opened 2024-04-23
Remove payloadStore

NiloCK:
PR replaces the engine's `payloadStore` with a single pointer that
- gets written to during calls to `forkChoiceUpdated`, "kicking off" a build process
- is referenced for the correct ID during calls to `getPayload`
- is used to construct the returned PayloadEnvelope
- is then discarded

PR replaces #8. Removal is the greatest refactor.

## Summary by CodeRabbit

- **New Features**
  - Introduced new HTTP service and node package for managing blockchain network operations.
  - Added new types for transaction transformations.

- **Improvements**
  - Updated engine API to handle payload attributes directly, enhancing performance and simplifying the codebase.
  - Enhanced test coverage configurations and Makefile to improve development practices.

- **Documentation**
  - Updated README.md with new instructions for running unit tests and generating coverage profiles.

- **Refactoring**
  - Renamed and adjusted structures and references in multiple files to align with new architectural decisions.
  - Improved code style in various configuration and test files for better readability and maintenance.


engine/engine.go
```diff
@@ -45,11 +44,11 @@ type TxValidator interface {
...
 	return &EngineAPI{
-		txValidator:  txValidator,
-		blockStore:   blockStore,
-		builder:      b,
-		payloadStore: payloadstore.NewPayloadStore(),
-		adapter:      adapter,
+		txValidator: txValidator,
+		blockStore:  blockStore,
+		builder:     b,
+		payloadAttr: nil,
```
joshklop, line 50: We don't need to explicitly set the field to it's zero value.

engine/engine.go
```diff
@@ -199,7 +195,7 @@ func (e *EngineAPI) ForkchoiceUpdatedV3(
 		}
 	}
 
-	payload := &monomer.Payload{
+	mpa := &monomer.PayloadAttributes{
```
joshklop, line 198:
```suggestion
	e.payloadAttr = &monomer.PayloadAttributes{
```

engine/engine.go
```diff
@@ -241,24 +238,26 @@ func (e *EngineAPI) GetPayloadV3(payloadID engine.PayloadID) (*eth.ExecutionPayl
...
-		NoTxPool:     payload.NoTxPool,
+		Transactions: e.payloadAttr.CosmosTxs,
+		GasLimit:     e.payloadAttr.GasLimit,
+		Timestamp:    e.payloadAttr.Timestamp,
+		NoTxPool:     e.payloadAttr.NoTxPool,
 	}); err != nil {
 		log.Panicf("failed to commit block: %v", err) // TODO error handling. An error here is potentially a big problem.
 	}
-	// TODO remove payload from payload store.
+	payloadEnvelope := e.payloadAttr.ToExecutionPayloadEnvelope(e.blockStore.HeadBlock().Hash())
```
joshklop, line 255: Do we need this variable or can we do 
```go
	return payload.ToExecutionPayloadEnvelope(e.blockStore.HeadBlock().Hash()), nil
```
as before?

NiloCK: We need to store this temporarily in order to set the payloadAttr to nil before `return`ing. Probably less awkward after the payloadenvelope construction is fixed to improve all block data.

engine/engine.go
NiloCK: :+1:

4b63610

engine/engine.go
NiloCK: 6fe0708

engine/engine.go
```diff
@@ -25,12 +24,12 @@ type BlockStore interface {
...
-	builder      *builder.Builder
-	txValidator  TxValidator
-	blockStore   BlockStore
-	payloadStore payloadstore.PayloadStore
-	adapter      PayloadTxAdapter
-	lock         sync.RWMutex
+	builder     *builder.Builder
+	txValidator TxValidator
+	blockStore  BlockStore
+	payloadAttr *monomer.PayloadAttributes
```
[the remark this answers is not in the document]
NiloCK: I like the addition of `current` more than the expansion of Attr (just for line length purposes), but I did them both :)

f5b9b57

engine/engine.go
```diff
@@ -233,24 +227,30 @@ func (e *EngineAPI) GetPayloadV3(payloadID engine.PayloadID) (*eth.ExecutionPayl
 	e.lock.RLock()
 	defer e.lock.RUnlock()
 
-	payload := e.payloadStore.Current()
-	if currentID := payload.ID(); payloadID != *currentID {
+	if e.currentPayloadAttributes == nil {
+		return nil, engine.InvalidParams.With(errors.New("engine payload is nil"))
```
joshklop, line 231:
```suggestion
		return nil, engine.InvalidParams.With(errors.New("payload not found"))
```
nit: error message that may make more sense to the user, and indicates that we haven't receive the attributes from forkchoiceUpdated

engine/engine.go
```diff
@@ -241,24 +238,26 @@ func (e *EngineAPI) GetPayloadV3(payloadID engine.PayloadID) (*eth.ExecutionPayl
 	e.lock.RLock()
 	defer e.lock.RUnlock()
 
-	payload := e.payloadStore.Current()
-	if currentID := payload.ID(); payloadID != *currentID {
+	if payloadID != *e.payloadAttr.ID() {
```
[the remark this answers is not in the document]
joshklop: I think explicitly checking nil is fine here. I agree the safest approach is to check for nil internally in general, but because the engine is the one setting the value to nil, I believe it bears responsibility here.

engine/engine.go
```diff
@@ -112,9 +110,6 @@ func (e *EngineAPI) ForkchoiceUpdatedV3(
 		if err := e.builder.Rollback(fcs.HeadBlockHash, fcs.SafeBlockHash, fcs.FinalizedBlockHash); err != nil {
 			return nil, engine.GenericServerError.With(fmt.Errorf("rollback: %v", err))
 		}
-		if err := e.payloadStore.RollbackToHeight(headBlock.Header.Height); err != nil {
```
joshklop, line 115: Should we set the payload attrs to `nil` here?

monomer.go
```diff
@@ -201,7 +201,7 @@ func hashDataAsBinary(h hash.Hash, data any) {
 }
 
 // ToExecutionPayloadEnvelope converts a Payload to an ExecutionPayload.
-func (p *Payload) ToExecutionPayloadEnvelope(blockHash common.Hash) *opeth.ExecutionPayloadEnvelope {
+func (p *PayloadAttributes) ToExecutionPayloadEnvelope(blockHash common.Hash) *opeth.ExecutionPayloadEnvelope {
```
joshklop, line 204: Comment above should be deleted, as it's now out of date. I don't think it needs to be modified, as the function is pretty obvious.

joshklop (approved):
Love it

monomer.go
NiloCK: 7b28c2e

engine/engine.go
NiloCK: I don't think so.
 - `currentPayloadAttributes` ought to already be `nil`
 - the currently running function is carrying a `payloadAttributes` (`pa`) that will be used soon to populate it
 
The removed rollback call was presumably in place for circumstances where more involved payloadAttributes storage was being anticipated.

PR #18, opened 2024-04-24
Refactor rollup type adapters

joshklop:
Not sure if this is useful or correct. I initially did this while implementing the e2e tests.

Even if we don't merge this, these changes can serve as inspiration for your PRs @NiloCK


joshklop:
Not planned. Was mainly used for reference.

Issue #17, opened 2024-04-23
Verify correct payloads returned via engineAPI


NiloCK:
Monomer's `Payload` type has a helper to construct `ExecutionPayload`s for consumption over the engineAPI.

When writing the executionPayload for a block, the helper uses the `payload'`s [transaction list](https://github.com/polymerdao/monomer/blob/8c335c273c578e04f443f6dbc9ee7dec8b803f47/monomer.go#L202).

This is potentially misleading / inaccurate, because the block itself is built with transactions sourced from this payload object [and also the mempool](https://github.com/polymerdao/monomer/blob/8c335c273c578e04f443f6dbc9ee7dec8b803f47/builder/builder.go#L97).

Should the returned ExecutionPayload be including this mempool tx data?

PR #20, opened 2024-04-30
Return all transactions with ExecutionPayloadEnvelope

NiloCK:
PR addresses issue #17.

Previously, only the transactions that had arrived via the engineAPI's `forkChoiceUpdated()` `payloadAttributes` parameter were being reported back in the execution payload. Appchain transactions sourced from the mempool were dropped.

PR modifies `monomer.PayloadAttributes.ToExecutionPayloadEnvelope()` to receive an entire `monomer.Block`, rather than just the hash.

## Summary by CodeRabbit

- **New Features**
  - Enhanced transaction handling by introducing `InjectedTransactions` for improved transaction injection and processing.

- **Bug Fixes**
  - Adjusted error handling in the building process to ensure better stability and clearer error messages.

- **Refactor**
  - Updated method signatures to return a block reference along with errors, facilitating better integration and error handling.
  - Modified payload struct and method parameters across multiple components to align with the new transaction handling approach.

- **Documentation**
  - Updated internal documentation to reflect changes in transaction processing and method outputs.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -85,8 +85,9 @@ func (b *Builder) Rollback(head, safe, finalized common.Hash) error {
 }
 
 type Payload struct {
-	// Transactions functions as an inclusion list.
-	Transactions bfttypes.Txs
+	// InjectedTransactions functions as an inclusion list. It contains transactions
+	// from the consensus layer that must be included in the block.
+	InjectedTransactions bfttypes.Txs
 	// TODO: make the gas limit actually be enforced. Need to translate between cosmos and op gas limit.
 	GasLimit  uint64
 	Timestamp uint64
@@ -94,7 +95,7 @@ type Payload struct {
 }
 
 func (b *Builder) Build(payload *Payload) error {
-	txs := slices.Clone(payload.Transactions) // Shallow clone is ok, we just don't want to modify the slice itself.
+	txs := slices.Clone(payload.InjectedTransactions) // Shallow clone is ok, we just don't want to modify the slice itself.
 	if !payload.NoTxPool {
 		for {
 			// TODO there is risk of losing txs if mempool db fails.
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -156,11 +156,13 @@ func (b *Builder) Build(payload *Payload) (*monomer.Block, error) {
 	})
 	b.app.Commit()
 
-	// Append block.
-	b.blockStore.AddBlock(&monomer.Block{
+	block := &monomer.Block{
 		Header: header,
 		Txs:    txs,
-	})
+	}
+
+	// Append block.
+	b.blockStore.AddBlock(block)
 	// Index txs.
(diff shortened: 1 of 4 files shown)


builder/builder.go
```diff
@@ -163,15 +163,15 @@ func (b *Builder) Build(payload *Payload) error {
...
 	for _, txResult := range txResults {
 		if err := b.eventBus.PublishEventTx(bfttypes.EventDataTx{
 			TxResult: *txResult,
 		}); err != nil {
-			return fmt.Errorf("publish event tx: %v", err)
+			return nil, fmt.Errorf("publish event tx: %v", err)
 		}
 	}
-	return nil
+	return b.blockStore.HeadBlock(), nil
```
joshklop, line 176: Let's build the struct here to avoid hitting the db.

builder/builder.go
```diff
@@ -94,7 +94,7 @@ type Payload struct {
 	NoTxPool  bool
 }
 
-func (b *Builder) Build(payload *Payload) error {
+func (b *Builder) Build(payload *Payload) (*monomer.Block, error) {
```
joshklop, line 97: There are some places in tests (maybe they're only in `builder_test.go`) where we call `HeadBlock` right after `Build`. We can clean up those call sites now that we return the block.

monomer.go
```diff
@@ -200,17 +200,23 @@ func hashDataAsBinary(h hash.Hash, data any) {
 	}
 }
 
-func (p *PayloadAttributes) ToExecutionPayloadEnvelope(blockHash common.Hash) *opeth.ExecutionPayloadEnvelope {
+func (p *PayloadAttributes) ToExecutionPayloadEnvelope(block Block) *opeth.ExecutionPayloadEnvelope {
```
joshklop, line 203: `*Block`

engine/engine.go
```diff
@@ -237,15 +237,17 @@ func (e *EngineAPI) GetPayloadV3(payloadID engine.PayloadID) (*eth.ExecutionPayl
 
 	// TODO: handle time slot based block production
 	// for now assume block is sealed by this call
-	if err := e.builder.Build(&builder.Payload{
+	head, err := e.builder.Build(&builder.Payload{
```
joshklop, line 240: `block` instead of `head`

joshklop (approved):

builder/builder.go
NiloCK: bd8ce7d

engine/engine.go
NiloCK: 9e9cf88

monomer.go
NiloCK: 8f886c8

builder/builder.go
NiloCK: 194b0f8

NiloCK:
Closing. Superseded by #24 (a modification of this PR that passes the e2e tests).

PR #24, opened 2024-05-07
Build ExecutionPayload from block data

NiloCK:
PR is a modification to #20, with fixes for the broken e2e test.

Current PR has the engine taking both adapters, and using the Cosmos->Eth adapter to convert mempool TXs from constructed blocks when returning the payloadEnvelope.

Notably: this conversion never gets invoked during the e2e, because there are no cosmos transactions processed.

PR has some merge conflicts. Open as draft to clean them up.

## Summary by CodeRabbit

- **New Features**
	- Introduced `DuplexAdapter` for handling Ethereum to Cosmos and vice versa transaction adaptations.
	- Added methods to convert and manage Ethereum and Cosmos transactions within bundles.
- **Refactor**
	- Updated transaction handling in the build process to use `InjectedTransactions`.
	- Enhanced error handling to return both block and error.
- **Bug Fixes**
	- Improved transaction adaptation logic in the engine and node components to ensure accurate processing.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -85,8 +85,9 @@ func (b *Builder) Rollback(head, safe, finalized common.Hash) error {
 }
 
 type Payload struct {
-	// Transactions functions as an inclusion list.
-	Transactions bfttypes.Txs
+	// InjectedTransactions functions as an inclusion list. It contains transactions
+	// from the consensus layer that must be included in the block.
+	InjectedTransactions bfttypes.Txs
 	// TODO: make the gas limit actually be enforced. Need to translate between cosmos and op gas limit.
 	GasLimit  uint64
 	Timestamp uint64
@@ -94,7 +95,7 @@ type Payload struct {
 }
 
 func (b *Builder) Build(payload *Payload) error {
-	txs := slices.Clone(payload.Transactions) // Shallow clone is ok, we just don't want to modify the slice itself.
+	txs := slices.Clone(payload.InjectedTransactions) // Shallow clone is ok, we just don't want to modify the slice itself.
 	if !payload.NoTxPool {
 		for {
 			// TODO there is risk of losing txs if mempool db fails.
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -176,6 +176,6 @@ func (b *Builder) Build(payload *Payload) (*monomer.Block, error) {
 		}
 	}
 
-  // TODO publish other things like new blocks.
+	// TODO publish other things like new blocks.
 	return block, nil
 }
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -156,11 +156,13 @@ func (b *Builder) Build(payload *Payload) (*monomer.Block, error) {
 	})
 	b.app.Commit()
 
-	// Append block.
-	b.blockStore.AddBlock(&monomer.Block{
+	block := &monomer.Block{
 		Header: header,
 		Txs:    txs,
-	})
+	}
+
+	// Append block.
+	b.blockStore.AddBlock(block)
 	// Index txs.
 	if err := b.txStore.Add(txResults); err != nil {
 		return nil, fmt.Errorf("add tx results: %v", err)
@@ -173,5 +175,5 @@ func (b *Builder) Build(payload *Payload) (*monomer.Block, error) {
 			return nil, fmt.Errorf("publish event tx: %v", err)
 		}
 	}
-	return b.blockStore.HeadBlock(), nil
+	return block, nil
 }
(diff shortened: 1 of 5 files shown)


monomer.go
```diff
@@ -38,6 +38,11 @@ type Application interface {
 	RollbackToHeight(uint64) error
 }
 
+type DuplexAdapter struct {
+	CosmosToEth CosmosTxAdapter
+	EthToCosmos PayloadTxAdapter
+}
+
```
joshklop, line 45: Is this a convenience type? I'm not understanding its purpose.

engine/engine.go
```diff
@@ -237,16 +238,44 @@ func (e *EngineAPI) GetPayloadV3(payloadID engine.PayloadID) (*eth.ExecutionPayl
...
+	}
+
+	txBytes := make([]hexutil.Bytes, len(txs))
+	for i, tx := range txs {
+		// return deposit transactions directly from the received payload attributes
+		if i < len(e.currentPayloadAttributes.Transactions) {
+			txBytes[i] = e.currentPayloadAttributes.Transactions[i]
+		} else {
+			txBytes[i] = tx.Data()
+		}
```
joshklop, line 263:
```suggestion
		txBytes[i] = tx.Data()
```
Any reason we can do this?

monomer.go
NiloCK: Just convenience - grouping similar functions.

engine/engine.go
NiloCK: That was the prior attempt - doesn't work because of some mismatched data (types?).

See https://www.notion.so/nethermind/36d9017dfafc440e96a8616b269bf619?v=90d07fe7c34e40109dd229c7605d7ca1&p=f071819347f048afa1cfa6c9c86d4c06&pm=s for some logged details from e2e tests.

I expect this to get resolved as the adapters become better tested.

monomer.go
joshklop: Personally, I'd prefer either putting them as receivers on the same type or leaving them entirely separate. Having functions as fields on a struct is generally confusing imho (the `SelectiveListener`s being the only useful case I can think of).

NiloCK: Removed the grouping struct here: be386bb

builder/builder.go
```diff
@@ -163,15 +163,15 @@ func (b *Builder) Build(payload *Payload) error {
...
 	for _, txResult := range txResults {
 		if err := b.eventBus.PublishEventTx(bfttypes.EventDataTx{
 			TxResult: *txResult,
 		}); err != nil {
-			return fmt.Errorf("publish event tx: %v", err)
+			return nil, fmt.Errorf("publish event tx: %v", err)
 		}
 	}
-	return nil
+	return b.blockStore.HeadBlock(), nil
```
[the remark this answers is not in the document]
NiloCK: This is already done in the PR - see 194b0f8f5f0ade7b13 - I guess your comment here was an an individual commit view?

engine/engine.go
joshklop: Ah, I see what I did wrong. `tx.Data()` only returns the data field of a deposit tx; it doesn't marshal the tx itself.

This should work:

```go
		txBytes[i], err = tx.MarshalBinary()
		if err != nil {
			return nil, engine.GenericServerError.With(fmt.Errorf("marshal tx binary: %v", err))
		}
```

builder/builder.go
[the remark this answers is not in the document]
joshklop: Yep, my bad

engine/engine.go
NiloCK: Oh, great. I had used another marshalling function unsuccessfully (ie, it failed the tests) and maybe confused it with this function.

ccef2f1

builder/builder_test.go
```diff
@@ -114,13 +114,14 @@ func TestBuild(t *testing.T) {
...
-				Timestamp:    g.Time + 1,
-				NoTxPool:     test.noTxPool,
+				InjectedTransactions: bfttypes.ToTxs(inclusionListTxs),
+				GasLimit:             0,
+				Timestamp:            g.Time + 1,
+				NoTxPool:             test.noTxPool,
 			}
 			preBuildInfo := app.Info(abcitypes.RequestInfo{})
-			require.NoError(t, b.Build(payload))
+			_, err = b.Build(payload)
```
joshklop, line 123: We should add a test ensuring that the block returned here matches `wantBlock` below.

`require.Equal(t, wantBlock, builtBlock)` or something.

joshklop (commented):
Looking good!

builder/builder_test.go
NiloCK: Good catch. 216237b

PR #26, opened 2024-05-08
Remove race condition in builder test

joshklop:
It was possible that the event bus subscription channel had not received all of the events by the time we were requiring it. Removing the check ensures we wait for the subscription channel to receive each event.

I'm not sure how we can ensure additional message aren't sent on the subscription. That is left untested, as before.

### DIFF
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -158,7 +158,6 @@ func TestBuild(t *testing.T) {
 
 			// Tx store and event bus.
 			eventChan := subscription.Out()
-			require.Len(t, eventChan, len(wantBlock.Txs))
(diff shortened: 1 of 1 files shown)


NiloCK (approved):
:+1: 

Could `require.Len(t, eventChan, 0)` after the loop to ensure there were no surprises, but seems pretty unlikely that it could happen.

joshklop:
> Could require.Len(t, eventChan, 0) after the loop to ensure there were no surprises, but seems pretty unlikely that it could happen.

Yeah an event could arrive right after that check passes, maybe giving some false sense of security. I'll leave it as-is for now.

PR #28, opened 2024-05-15
ABCI 2.0

joshklop:


testapp/testapp.go
NiloCK, line 1: This is a mouthful, and the specifics of the app construction via the `Inject` are pretty obscure to me, but it all reads fine.

NiloCK: Helps that the large change is against small changes to `testapp_test.go`

testutils/utils.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func NewCometMemDB(t *testing.T) cometdb.DB {
+	db := cometdb.NewMemDB()
+	t.Cleanup(func() {
+		require.NoError(t, db.Close())
+	})
+	return db
+}
```
NiloCK, line 17: Again unsure about why this was introduced.

x/rollup/types/keys.go
```diff
@@ -7,9 +7,6 @@ const (
 	// StoreKey defines the primary module store key
 	StoreKey = ModuleName
 
-	// RouterKey is the message route for slashing
-	RouterKey = ModuleName
```
NiloCK, line 11: Do I read correctly here that all these consts are `"rollup"`? Is this placeholder?

gen/rollup/v1/helpers.go
```diff
@@ -0,0 +1,34 @@
...
+	return nil
+}
+
+func (*ApplyL1TxsRequest) Type() string {
+	return "l1txs"
+}
+
+func (*ApplyL1TxsRequest) Route() string {
+	return "rollup"
+}
```
NiloCK, line 30: I think these are unused?

NiloCK (approved):
This looks good to me, but my review is not exhaustive.

In future we should try to isolate major dependency changes from other functional changes to make for easier digesting PRs.

proto/buf.yaml
```diff
@@ -0,0 +1,11 @@
+version: v1
+deps:
+  - buf.build/cosmos/gogo-proto
```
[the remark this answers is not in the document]
joshklop: Yes this is a cosmos-specific fork that we have to use. I think they want to move off of it, but the whole ecosystem is stuck with it for now.

testapp/proto/testapp/v1/set.proto
```diff
@@ -2,9 +2,14 @@ syntax = "proto3";
...
+import "cosmos/msg/v1/msg.proto";
+
 message SetRequest {
-    string key = 1;
-    string value = 2;
+    option (cosmos.msg.v1.signer) = "from_address";
+
+    string from_address = 1;
+    string key = 2;
+    string value = 3;
```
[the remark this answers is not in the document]
joshklop: There is `buf breaking`, which is supposed to catch breaking changes. I'm not worrying about that for now because we are still early in development.

testapp/testapp.go
joshklop: > the specifics of the app construction via the Inject are pretty obscure to me, but it all reads fine.

Same for me. I don't understand how it all works.

gen/rollup/v1/helpers.go
joshklop: Correct. They are used to fulfill interfaces that the cosmos-sdk requires.

node/node.go
```diff
@@ -39,9 +40,9 @@ type Node struct {
 	engineHTTP                 net.Listener
 	engineWS                   net.Listener
 	cometHTTPAndWS             net.Listener
-	blockdb                    tmdb.DB
-	txdb                       tmdb.DB
-	mempooldb                  tmdb.DB
+	blockdb                    dbm.DB
+	txdb                       cometdb.DB
+	mempooldb                  dbm.DB
```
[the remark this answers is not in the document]
joshklop: The cosmos sdk migrated to the new db package. The `txdb` needs to still be the comet db because we use comet's indexing functionality.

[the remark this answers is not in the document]
joshklop: I agree it's messy. We have a to-do to fix the db layer on the task board.

joshklop:
> In future we should try to isolate major dependency changes from other functional changes to make for easier digesting PRs.

I agree. I unsuccessfully attempted to break some of this into smaller pieces, but the dependencies kept conflicting. Upgrading them all at once appeared to be the only option.

testutils/utils.go
joshklop: Good point. Answered above.

PR #27, opened 2024-05-10
Add cosmos Tx consumption / rejection to e2e

NiloCK:
PR
- submits a cosmos TX
- asserts that the Tx is retrievable
- attempts to submit a malformed Tx, asserts that the submission fails

Also: changes the RPC server to reject transactions that fail `CheckTx`, instead of feeding them to the mempool (where they're destined to be rejected later).

## Summary by CodeRabbit

- **New Features**
	- Enhanced transaction handling to improve reliability in processing user transactions.

- **Bug Fixes**
	- Adjusted transaction enqueuing logic to ensure only valid transactions are processed, enhancing system stability.

- **Tests**
	- Expanded test coverage to include new scenarios for transaction broadcasting and result handling, ensuring robustness and reliability.

PR #29, opened 2024-05-16
Add regression test for fail-fast on bad CosmosTx

NiloCK:
PR adds a failure case to `TestBroadcastTx`:
- create a malformed TX
- submit it
- asserts that it is not enqueued by the mempool

The updated test was manually run against the prior implementation of `BroadcastTx`, and failed. The revised, current, `BroadcastTx` (see #27) passes.

### DIFF
--- a/comet/comet_test.go
+++ b/comet/comet_test.go
@@ -126,6 +126,8 @@ func TestBroadcastTx(t *testing.T) {
 	app := testapp.NewTest(t, chainID)
 	mpool := mempool.New(testutil.NewMemDB(t))
 	broadcastAPI := comet.NewBroadcastTx(app, mpool)
+
+	// Succuss case.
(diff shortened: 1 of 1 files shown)


comet/comet_test.go
```diff
@@ -137,6 +139,21 @@ func TestBroadcastTx(t *testing.T) {
...
+
+	// Error case - malformed tx.
+	badTx := []byte{1, 2, 3}
+	startLen, err := mpool.Len()
+	require.NoError(t, err)
+
+	result, err = broadcastAPI.BroadcastTx(nil, badTx)
+	// API does not error, but returns a non-zero error code.
+	require.NoError(t, err)
+	require.NotEqual(t, uint32(0), result.Code)
```
joshklop, line 151: `require.True(t, result.IsErr()`

Does this work?

joshklop (approved):

PR #32, opened 2024-05-28
Improve logging in e2e tests

joshklop:


NiloCK (approved):
This LGTM with a couple of loose suggestions:

- timestamping the process outputs is useful when cross-referencing
- maybe: collapse the individual process loggers into a single function

joshklop:
> timestamping the process outputs is useful when cross-referencing

This is useful, though I'd prefer to do this using external command line tools to minimize code in Monomer.

```bash
# something like this should do the trick:
multitail -ts -i e2e/artifacts/anvil.log -ts -i e2e/artifacts/op.log
```

> collapse the individual process loggers into a single function

Good idea. Will do, thanks!

PR #35, opened 2024-06-08
Inline tx adapters

phasewalk1:
This PR "inlines" the transaction adapters to and from Cosmos SDK and Ethereum types. Previously, they were being passed around as callbacks (we had needed this abstraction to avoid a dependency cycle, but that's no longer a concern). 

Essentially, we are removing the callbacks and instead calling the adapter implementations directly, for example, [`AdaptCosmosTxsToEthTxs`](https://github.com/polymerdao/monomer/blob/ec97025ded408a1ddd92619bccbec3bc0f9156a3/x/rollup/types/adapters.go#L62).

Requesting review: @joshklop @NiloCK 

## Summary by CodeRabbit

- **Refactor**
  - Simplified transaction adaptation by removing redundant adapter functions and fields across multiple components.
  - Updated function signatures and internal calls to use streamlined transaction adaptation methods.

- **Tests**
  - Simplified test functions by removing unnecessary parameters related to transaction adaptation.

### DIFF
--- a/e2e/stack.go
+++ b/e2e/stack.go
@@ -23,7 +23,6 @@ import (
 	"github.com/polymerdao/monomer/genesis"
 	"github.com/polymerdao/monomer/node"
 	"github.com/polymerdao/monomer/testapp"
-	rolluptypes "github.com/polymerdao/monomer/x/rollup/types"
 )
 
 const oneETH = uint64(1e18)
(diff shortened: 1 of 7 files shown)


joshklop:
Can we remove the adapter types in the top-level `monomer` package?

phasewalk1:
> Can we remove the adapter types in the top-level `monomer` package?

Yes, I don't believe they are used anywhere. I'll do that now.

joshklop (approved):
LGTM. Good to merge after we figure out these CI checks.

PR #36, opened 2024-06-10
Avoid overwriting log record attributes in e2e

joshklop:
See commit message

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved attribute handling to prevent overwriting existing attributes and ensure accumulation of attributes.


NiloCK (approved):
This is a fix, but the function is unused as far as I can tell.

?

joshklop:
> This is a fix, but the function is unused as far as I can tell.
> 
> ?

It fulfills an interface (`slog.Handler` I think it is) that is used by by the `oplog` package.

PR #37, opened 2024-06-10
Set tx sender address in config so signatures succeed

joshklop:
Should work for both the batcher and proposer.

## Summary by CodeRabbit

- **New Features**
  - Enhanced security by initializing configurations with a public key address.


NiloCK:
Interesting that this is triggering the same transactions root calculation failure loop.

joshklop:
> Interesting that this is triggering the same transactions root calculation failure loop.

Yes I think it's because the batcher is now working properly (i.e., posting L2 data to L1 as it should). When the op-node reads the batches on L1, it hits the receipt root anvil issue.

NiloCK (approved):
Seems like an obviously correct insertion, but probably better to hold off merge until the e2e error is resolved.

joshklop:
Fixed in #68

PR #40, opened 2024-06-10
Upgrade go lint action

joshklop:
Caching was broken on the previous version.

## Summary by CodeRabbit

- **Chores**
  - Updated the version of `golangci/golangci-lint-action` to `v6` for improved linting performance and compatibility.


NiloCK (approved):
I don't 100% understand the link between the cache directive and the lint upgrade, but the existing run looks successful in any case.

:+1:

PR #42, opened 2024-06-12
Cosmos SDK Integration

phasewalk1:
## Scope
This PR is the first step of iteration in integrating the Cosmos SDK with Monomer. This will allow Cosmos developers to easily deploy their applications as Ethereum rollups.

The Cosmos SDK (since ~`v0.50.6`) exposes a `StartCommandHandler` callback that overrides the default `start` function in [`server/start.go`](https://github.com/cosmos/cosmos-sdk/blob/74ae427575a54d13f9065b00995f34175fd39c2e/server/start.go). We customize this to slot in an in-process Monomer node instead of a CometBFT node.

Since this PR is just the first step of iteration, we've simplified a lot of complexity that is present in the Cosmos SDK. For now, we use a [`fakeTraceWriter`](https://github.com/phasewalk1/monomer/blob/8ef837d6c396f4e988b4ffddd8570a8b23224ce2/integrations/integrations.go#L74), we ignore things like telemetry and server flags (like `IsTestnet`, `gRPCOnly`, etc.), and we've redesigned some of the helper functions carried over from the SDK to work better for our use cases.

## Summary by CodeRabbit

- **New Features**
  - Introduced a custom callback to start a Monomer node in-process, enhancing integration with Cosmos SDK applications.
  - Implemented new functions to manage application startup, databases, and network listeners.
- **Tests**
  - Added new test functions to ensure robustness of the Monomer node integration with Cosmos SDK.
  - Provided a structured configuration for testing various modules and settings.


integrations/integrations.go
```diff
@@ -0,0 +1,201 @@
...
+	cometListener, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		return nil, nil, err
+	}
+	appdb := dbm.NewMemDB()
+
+	// We obviously need to replace this with an actual Monomer Application, not
+	// a testapp.App. How can I do this?
+	app, err := testapp.New(appdb, chainID.HexBig().String())
+	if err != nil {
```
joshklop, line 152: We should be able to use the `servertypes.Application` that's in `startInProcess`. We'll probably need to wrap it to implement the `RollbackToHeight` function.

integrations/integrations.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// Starts the Monomer node in-process in place of the Comet node.
+func startMonomerNode() (*node.Node, func(), error) {
+	chainID := monomer.ChainID(1)
+	engineWS, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		return nil, nil, err
+	}
+	cometListener, err := net.Listen("tcp", "127.0.0.1:0")
```
joshklop, line 143: We should obtain the addresses from the config. We may need to add some configurations (or just use environment variables if the Cosmos SDK doesn't give us hooks to add configs).

integrations/integrations.go
```diff
@@ -0,0 +1,201 @@
...
+
+	g, ctx := GetCtx(svrCtx, true)
+
+	if gRPCOnly {
+		svrCtx.Logger.Info("starting node in gRPC only mode; Monomer is disabled")
+		svrCfg.GRPC.Enable = true
+	} else {
+		svrCtx.Logger.Info("starting Monomer node in-process")
+		// Start the Monomer node
+		_, monomerCleanupFn, err := startMonomerNode()
```
joshklop, line 100: We can pass in an `environment.Env` here to avoid dealing with the awkward cleanup function.

integrations/integrations.go
```diff
@@ -0,0 +1,201 @@
...
+			OnEngineWebsocketServeErrCb: func(err error) {
+				fmt.Errorf("failed to serve engine websocket: %w", err)
+			},
+		},
+	)
+
+	nodeCtx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	env := environment.New()
```
joshklop, line 195: Need to call `env.Close`. We should pass in the environment to this function and call `env.Defer`/`env.DeferErr` instead of having a cleanup function.

phasewalk1: So we call `env.Close` after doing the cleanup in `env.Defer`?

integrations/integrations.go
phasewalk1: I'm a little confused, could you clarify?

I see the `RollbackToHeight` implementation we have in `testapp.go` [here](https://github.com/polymerdao/monomer/blob/81c3603e87322133c9cc73cc5368c816060136bb/testapp/testapp.go#L83). But how do I _wrap_ the `servertypes.Application` to implement this?

joshklop: We need access to the `CommitMultiStore` to rollback (the link above demonstrates that). We may need an ugly type assertion to get access to it... but it might be available some other way.

integrations/integrations.go
phasewalk1: Addressed.

integrations/integrations.go
phasewalk1: Addressed.

integrations/integrations.go
phasewalk1: Addressed by creating a wrapper around `servertypes.Application` that reconciles the discrepancies between it and the `monomer.Application` interface.

integrations/integrations.go
phasewalk1: Got it. Do we want the P2P address or the RPC?

joshklop: RPC. We don't do p2p right now.

phasewalk1: Addressed.

NiloCK (commented):
Looks great. Will leave approval for @joshklop, since he's got a much better picture of the overall goals here.

integrations/helpers.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+
+	"github.com/cosmos/cosmos-sdk/server"
+	serverconfig "github.com/cosmos/cosmos-sdk/server/config"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	FlagTraceStore = "trace-store"
+	FlagGRPCOnly   = "grpc-only"
```
joshklop, line 13: needed?

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+	// We assume `inProcessConsensus` is true for now, so let's return an error if it's not.
+	if !inProcessConsesus {
+		return fmt.Errorf("in-process consensus must be enabled")
+	}
+
+	env := environment.New()
+	defer func() {
+		if closeErr := env.Close(); closeErr != nil {
+			svrCtx.Logger.Error(closeErr.Error())
+		}
```
joshklop, line 44: nit: simpler names, capitalized log message
```suggestion
		if err := env.Close(); err != nil {
			svrCtx.Logger.Error("Failed to close environment", "err", err)
		}
```

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+	}()
+
+	// --- Do the normal Cosmos SDK stuff ---
+	svrCfg, err := getAndValidateConfig(svrCtx)
```
joshklop, line 48: Let's inline `getAndValidateConfig`

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+func startApp(
+	env *environment.Env,
+	svrCtx *server.Context,
+	appCreator servertypes.AppCreator,
+	opts server.StartCmdOptions,
+) (app servertypes.Application, err error) {
+	home := svrCtx.Config.RootDir
+	db, err := opts.DBOpener(home, server.GetAppDBBackend(svrCtx.Viper))
+	if err != nil {
+		return app, err
```
joshklop, line 94: almost always preferred to embellish error strings. Makes it easier to track bugs down
```suggestion
		return app, fmt.Errrof("open db: %v", err")
```

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+
+// See https://github.com/cosmos/cosmos-sdk/blob/7fb26685cd68a6c1d199dc270c80f49f2bfe7ace/server/start.go#L624
+func startApp(
+	env *environment.Env,
+	svrCtx *server.Context,
+	appCreator servertypes.AppCreator,
+	opts server.StartCmdOptions,
+) (app servertypes.Application, err error) {
+	home := svrCtx.Config.RootDir
+	db, err := opts.DBOpener(home, server.GetAppDBBackend(svrCtx.Viper))
```
joshklop, line 92:
```suggestion
	db, err := opts.DBOpener(svrCtx.Config.RootDir, server.GetAppDBBackend(svrCtx.Viper))
```

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return app, err
+	}
+
+	// TODO: Check if is testnet and implement `testnetify` function
```
joshklop, line 97: Is this our todo or from the SDK?

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+
+	return app, nil
+}
+
+// See https://github.com/cosmos/cosmos-sdk/blob/7fb26685cd68a6c1d199dc270c80f49f2bfe7ace/server/start.go#L307
+// We modify this function to start a Monomer node in-process instead of a Comet node.
+func startInProcess(
+	env *environment.Env,
+	svrCtx *server.Context,
+	_ serverconfig.Config, //nolint:gocritic // hugeParam
```
joshklop, line 114: can we remove the unused param?

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+	clientCtx client.Context, //nolint:gocritic // hugeParam
```
joshklop, line 115: Can we pass this as a pointer and then dereference later?

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+	app servertypes.Application,
+	opts server.StartCmdOptions,
+) error {
+	g, ctx := getCtx(svrCtx, true)
+
+	svrCtx.Logger.Info("starting Monomer node in-process")
+	// --- Start the Monomer node ---
```
joshklop, line 122: nit: we can remove this comment, the function call is pretty self-explanatory (similar for the other `--- ... ---` comments).

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	return g.Wait()
+}
+
+// Starts the Monomer node in-process in place of the Comet node. The
+// `MonomerGenesisPath` flag must be set in viper before invoking this.
+func startMonomerNode(wrappedApp *WrappedApplication, env *environment.Env, svrCtx *server.Context) (*node.Node, error) {
+	// --- Bootstrap listeners and DBs ---
+	chainID := monomer.ChainID(1)
```
joshklop, line 142: We can get the chain ID from `appGenesis.ChainID`. That way it's configurable.

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+	engineWS, err := net.Listen("tcp", "127.0.0.1:0")
```
joshklop, line 143: This address needs to be configurable. We may need `monomer-engine-ws-addr` or something.

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+	cmtListenAddr = strings.TrimPrefix(cmtListenAddr, "tcp://")
+	cometListener, err := net.Listen("tcp", cmtListenAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	appdb := dbm.NewMemDB()
+	blockdb := dbm.NewMemDB()
+	txdb := cometdb.NewMemDB()
+	mempooldb := dbm.NewMemDB()
```
joshklop, line 158: For the listeners and dbs, let's defer `Close` right away. This way the resources are closed if we hit a premature error or a panic

e.g.,
 ```suggestion
	appdb := dbm.NewMemDB()
	env.DeferErr("close app db", appdb.Close)
	blockdb := dbm.NewMemDB()
	env.DeferErr("close block db", blockdb.Close)
	txdb := cometdb.NewMemDB()
	env.DeferErr("close tx db", txdb.Close)
	mempooldb := dbm.NewMemDB()
	env.DeferErr("close mempool db", mempooldb.Close)
```

File: integrations/integrations.go [hunk 17]
```diff
@@ -0,0 +1,232 @@
...
+
+	// --- Load Monomer genesis doc ---
+	monomerGenesisPath := viper.GetString("monomer-genesis-path")
+	svrCtx.Logger.Info("loading Monomer genesis from", "path", monomerGenesisPath)
```
joshklop, line 162: nit: log messages should be capitalized
```suggestion
	svrCtx.Logger.Info("Loading Monomer genesis", "path", monomerGenesisPath)
```

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+
+	appGenesis, err := genutiltypes.AppGenesisFromFile(monomerGenesisPath)
+	if err != nil {
+		svrCtx.Logger.Error("failed to load Monomer genesis", "error", err)
+		return nil, err
```
joshklop, line 167: Prefer returning embellished errors over log messages, generally (same for other places in this file).
```suggestion
		return nil, fmt.Errorf("load application genesis file: %v", err)
```

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+// Starts the Monomer node in-process in place of the Comet node. The
+// `MonomerGenesisPath` flag must be set in viper before invoking this.
+func startMonomerNode(wrappedApp *WrappedApplication, env *environment.Env, svrCtx *server.Context) (*node.Node, error) {
+	// --- Bootstrap listeners and DBs ---
+	chainID := monomer.ChainID(1)
+	engineWS, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		return nil, err
```
joshklop, line 145: nit: embellishing errors (we should do the same for other errors that are returned)
```suggestion
		return nil, fmt.Errorf("create engine websocket listener: %v", err)
```

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	cmtListenAddr := svrCtx.Config.RPC.ListenAddress
+	cmtListenAddr = strings.TrimPrefix(cmtListenAddr, "tcp://")
+	cometListener, err := net.Listen("tcp", cmtListenAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	appdb := dbm.NewMemDB()
```
joshklop, line 155: Using only in-memory databases is fine for tests, but this code will run in production where data is usually persisted.

I think this can help us let the user specify the db backend, but let me know if you find a better way. https://github.com/cometbft/cometbft/blob/f4c622fe597f925df9d7cd2842b726eab915aa98/config/db.go#L25

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	// --- Check if CometBFT listener is up ---
+	if conn, err := net.Dial("tcp", cmtListenAddr); err != nil {
+		svrCtx.Logger.Error("failed to dial CometBFT", "error", err)
+	} else {
+		svrCtx.Logger.Info("dialing CometBFT succeeded at", "address", cmtListenAddr)
+		conn.Close()
+	}
+
```
joshklop, line 217: This is unnecessary, correct?

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+				svrCtx.Logger.Error("failed to serve engine websocket", "error", err)
+			},
+			OnCometServeErrCb: func(err error) {
+				svrCtx.Logger.Error("failed to serve CometBFT", "error", err)
+			},
+		},
+	)
+	svrCtx.Logger.Info("spinning up Monomer node")
+
+	nodeCtx, nodeCtxCancel := context.WithCancel(context.Background())
```
joshklop, line 202: Let's defer the cancel here instead of waiting til the end of the function
```suggestion
	nodeCtx, nodeCtxCancel := context.WithCancel(context.Background())
	env.Defer(nodeCtxCancel)
```

joshklop (commented):
Nice work! Left some comments, mostly small things.

integrations/helpers.go
phasewalk1: No longer needed. Removing

integrations/integrations.go
phasewalk1: I wrote this `TODO` based on the SDK function `startApp`. The line it's referencing is [here](https://github.com/cosmos/cosmos-sdk/blob/7fb26685cd68a6c1d199dc270c80f49f2bfe7ace/server/start.go#L636)

integrations/integrations.go
phasewalk1: yup! I had it there initially when copying over from the SDK. Removing now

integrations/integrations.go
phasewalk1: Pass `clientCtx` as a pointer?

integrations/integrations.go
phasewalk1:
```go
appGenesis, err := genutiltypes.AppGenesisFromFile(monomerGenesisPath)
if err != nil {
    svrCtx.Logger.Error("failed to load Monomer genesis", "error", err)
    return nil, err
}
```
`appGenesis.ChainID` is a string here. So I'd need to parse it into a `uint64` to use it as a `monomer.ChainID` type. Something like:
```go
appGenesis, err := genutiltypes.AppGenesisFromFile(monomerGenesisPath)
if err != nil {
    svrCtx.Logger.Error("failed to load Monomer genesis", "error", err)
    return nil, err
}
uintChainID, err := strconv.ParseUint(appGenesis.ChainID, 10, 64)
if err != nil {
    return nil, fmt.Errorf("failed to parse chain ID: %w", err)
}
chainID := monomer.ChainID(uintChainID)
```

integrations/integrations.go
phasewalk1: Acknowledged. Is there a good way to manage all of our viper fields? Since we also have `monomer-genesis-path`.

integrations/integrations.go
phasewalk1: Does `env.DeferErr("close app db", appdb.Close)` actually call `appdb.Close()`? To clarify, the `env.Defer` block at the end of `startMonomerNode` would be simplified to:
```go
env.Defer(func() {
    engineWS.Close()
    cometListener.Close()
    nodeCtxCancel()
})
```

integrations/integrations.go
phasewalk1: Yeah this is unnecessary. I added it for testing but will remove it now

integrations/helpers.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// See https://github.com/cosmos/cosmos-sdk/blob/7fb26685cd68a6c1d199dc270c80f49f2bfe7ace/server/start.go#L403
+func getAndValidateConfig(svrCtx *server.Context) (serverconfig.Config, error) {
+	config, err := serverconfig.GetConfig(svrCtx.Viper)
+	if err != nil {
+		return config, err
```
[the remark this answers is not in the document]
phasewalk1: Addressed. `getAndValidateConfig` is now inlined as well, as per @joshklop's feedback

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+		txdb,
+		&node.SelectiveListener{
+			OnEngineHTTPServeErrCb: func(err error) {
+				svrCtx.Logger.Error("failed to serve engine HTTP", "error", err)
+			},
+			OnEngineWebsocketServeErrCb: func(err error) {
+				svrCtx.Logger.Error("failed to serve engine websocket", "error", err)
+			},
+			OnCometServeErrCb: func(err error) {
+				svrCtx.Logger.Error("failed to serve CometBFT", "error", err)
```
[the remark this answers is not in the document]
phasewalk1: Good point. Addressed
```go
&node.SelectiveListener{
    OnEngineHTTPServeErrCb: func(err error) {
        svrCtx.Logger.Error("[Engine HTTP Server]", "error", err)
    },
    OnEngineWebsocketServeErrCb: func(err error) {
        svrCtx.Logger.Error("[Engine Websocket]", "error", err)
    },
    OnCometServeErrCb: func(err error) {
        svrCtx.Logger.Error("[CometBFT]", "error", err)
    },
},
```

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+		},
+	)
+	svrCtx.Logger.Info("spinning up Monomer node")
+
+	nodeCtx, nodeCtxCancel := context.WithCancel(context.Background())
+
+	// --- Run Monomer ---
+	nodeErr := n.Run(nodeCtx, env)
+	if nodeErr != nil {
+		svrCtx.Logger.Error("failed to run Monomer node", "error", err)
```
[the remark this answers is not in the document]
phasewalk1: Yes we should be returning an error here. Addressed!

integrations/testdata/genesis.json
```diff
@@ -0,0 +1,60 @@
...
+        "last_validator_powers": [],
+        "validators": [],
+        "delegations": [],
+        "unbonding_delegations": [],
+        "redelegations": [],
+        "exported": false
+      },
+      "testmodule": {}
+    },
+    "chain_id": "1",
```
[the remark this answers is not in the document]
phasewalk1: +1. As per @joshklop's feedback, I'm going to remove the line where `chainID` is hardcoded in `startMonomerNode` and instead, configure it with the app genesis file.

integrations/integrations_test.go
```diff
@@ -0,0 +1,187 @@
...
+		DBOpener: func(rootDir string, backendType dbm.BackendType) (dbm.DB, error) {
+			return dbm.NewMemDB(), nil
+		},
+	}
+
+	cmtListenAddr := svrCtx.Config.RPC.ListenAddress
+	cmtListenAddr = strings.TrimPrefix(cmtListenAddr, "tcp://")
+
+	go func() {
+		err := StartCommandHandler(svrCtx, clientCtx, mockAppCreator, inProcessConsensus, opts)
```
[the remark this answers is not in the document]
phasewalk1: This is a good question. The way I understand it is that we don't need them because our `environment.Env` handles cleanup. We also have to abide by the signature for `StartCommandHandler` in order to use it in the Cosmos SDK. Is this correct? cc @joshklop

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+	servertypes "github.com/cosmos/cosmos-sdk/server/types"
+	genutiltypes "github.com/cosmos/cosmos-sdk/x/genutil/types"
+	"github.com/polymerdao/monomer"
+	"github.com/polymerdao/monomer/environment"
+	"github.com/polymerdao/monomer/genesis"
+	"github.com/polymerdao/monomer/node"
+	"github.com/spf13/viper"
+)
+
+func StartCommandHandler(
```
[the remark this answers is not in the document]
phasewalk1: Agreed. Will add

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+	svrCtx *server.Context,
+	clientCtx client.Context, //nolint:gocritic // hugeParam
+	appCreator servertypes.AppCreator,
+	inProcessConsesus bool,
```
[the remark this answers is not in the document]
phasewalk1: Nice catch. Addressed

integrations/integrations.go
```diff
@@ -0,0 +1,232 @@
...
+		return nil, err
+	}
+
+	appdb := dbm.NewMemDB()
+	blockdb := dbm.NewMemDB()
+	txdb := cometdb.NewMemDB()
+	mempooldb := dbm.NewMemDB()
+
+	// --- Load Monomer genesis doc ---
+	monomerGenesisPath := viper.GetString("monomer-genesis-path")
```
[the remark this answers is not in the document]
phasewalk1: Agreed. I'm wondering what's the best way to manage our viper configs. Josh recommended we also set `monomer-engine-ws-addr` to configure the Engine websocket. Maybe something like this at the top of `integrations.go`?
```go
const (
    monomerGenesisPathFlag = "monomer-genesis-path"
    monomerEngineWSFlag    = "monomer-engine-ws"
)
```

integrations/integrations_test.go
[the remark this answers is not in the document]
joshklop: `StartCommandHandler` is kind of like a `main` function, so it is responsible for setup and shutdown.

I'm not saying I agree with this design, but as @phasewalk1 mentioned, we have to dance to the SDK's tune.

integrations/integrations.go
joshklop: Yes

integrations/integrations.go
joshklop: Parsing the string is ok.

integrations/integrations.go
joshklop: We shouldn't have too many. Separate string literals is fine with me. Colin's idea to use `const`s should suffice for now.

integrations/integrations.go
joshklop: > Does env.DeferErr("close app db", appdb.Close) actually call appdb.Close()

`appdb.Close()` is called in `env.Close()`.

> To clarify, the env.Defer block at the end of startMonomerNode would be simplified to:

Note that we want to defer cleanup as close to initialization as possible while capturing errors with `DeferErr`. So we should defer the `engineWS.Close()` closer to `engineWS := net.Listener(...)`. Same for the comet listener.

integrations/integrations.go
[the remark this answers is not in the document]
joshklop: LGTM. @NiloCK can you confirm that is what you were referring to?

integrations/integrations.go
phasewalk1: `DefaultDBProvider` seems like the best way to do this, but I'm not sure how to get the `DBContext` object required as an input.

joshklop: `DBContext` can be created manually from the `svrCtx.Config`, I believe (since all of its fields are exported).

integrations/integrations.go
```diff
@@ -0,0 +1,226 @@
...
+	env := environment.New()
+	defer func() {
+		if err := env.Close(); err != nil {
+			svrCtx.Logger.Error("Failed to close environment", "err", err)
+		}
+	}()
+
+	svrCfg, err := serverconfig.GetConfig(svrCtx.Viper)
+	if err != nil {
+		return fmt.Errorf("failed to get server config: %w", err)
```
joshklop, line 57: nit: embellished errors typically express the intent only. 
```suggestion
		return fmt.Errorf("get server config: %w", err)
```
By the time the error is printed to stdout, we don't want it to look like this:
```
error: failed to X: failed to Y: Z
```
We're shooting for:
```
error: X: Y: Z
```

phasewalk1: Ah I understand now, thanks for the clarification and for sharing the blog post. I'll check it out

integrations/integrations.go
phasewalk1: > `DBContext` can be created manually from the `svrCtx.Config`, I believe (since all of its fields are exported).

Cool so I'm thinking something like this:
```go
appdb, err := cometconfig.DefaultDBProvider(&cometconfig.DBContext{
    Config: svrCtx.Config,
})
```
I noticed `DBContext` also has an `ID` field ... Maybe something like `monomer-app-db`?

joshklop: > Maybe something like monomer-app-db?

Works for me. What does the `ID` field do?

integrations/integrations.go
joshklop: Got it, thanks! Mind adding this to the task board with a brief description as well? Feel free to resolve the comment once added.

integrations/integrations.go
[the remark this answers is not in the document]
NiloCK: Yup.

integrations/integrations.go
phasewalk1: The `ID` field is passed in as the `name` parameter to the `NewDB` call in `DefaultDBProvider`.

integrations/integrations.go
phasewalk1: Added to task board.

integrations/integrations.go
phasewalk1: I just noticed that using `DefaultDBProvider` creates a `integrations/data/` directory. I'm going to add it to the `.gitignore` @joshklop

integrations/integrations_test.go
```diff
@@ -0,0 +1,187 @@
...
+	// --- Ensure Monomer has an active CometBFT listener ---
+	if conn, err := net.Dial("tcp", cmtListenAddr); err != nil {
+		svrCtx.Logger.Error("failed to dial CometBFT", "error", err)
+		panic(err)
+	} else {
+		conn.Close()
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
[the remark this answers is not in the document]
phasewalk1: Good question. I'm still getting a hang for Go contexts so I shy away from answering definitively here. My intuition says that we don't need it because the only things we use it for are things we won't want to cancel, i.e., the only place it's actually passed in is to `bftClient.BroadcastTxAsync(ctx, txBytes)`. So `context.Background()` should suffice here.

integrations/integrations.go
phasewalk1: Should I do this for all DBs besides mempool?

joshklop: It is probably possible to change this using viper. We should change the viper config value in the tests to point to a `t.TemptDir` to avoid this.

phasewalk1: Addressed

integrations/integrations.go
```diff
@@ -0,0 +1,231 @@
...
+// See https://github.com/cosmos/cosmos-sdk/blob/7fb26685cd68a6c1d199dc270c80f49f2bfe7ace/server/start.go#L307
+// We modify this function to start a Monomer node in-process instead of a Comet node.
+func startInProcess(
+	env *environment.Env,
+	svrCtx *server.Context,
+	clientCtx *client.Context,
+	app servertypes.Application,
+	opts server.StartCmdOptions,
+) error {
+	g, ctx := getCtx(svrCtx, true)
```
joshklop, line 127: `g` and `ctx` are only used in `PostSetup`. Furthermore, `getCtx` makes us wait for signals again. Let's remove `getCtx` entirely and move something like this to lines 136-142:

```go
if opts.PostSetup != nil {
        g := errgroup.New()
        if err := opts.PostSetup(svrCtx, *clientCtx, ctx, g); err != nil {
                return fmt.Errorf("run post setup: %v", err)
        }
        return g.Wait()
}

return nil
``` 

where `ctx` is passed in by the caller.

integrations/integrations.go
```diff
@@ -0,0 +1,231 @@
...
+
+	svrCtx.Logger.Info("Starting Monomer node in-process")
+	wrappedApp := &WrappedApplication{app}
+	_, err := startMonomerNode(wrappedApp, env, svrCtx)
+	if err != nil {
```
joshklop, line 132: nit: limit the scope of variables.
```suggestion
	if _, err := startMonomerNode(&WrappedApplication{
	        app: app,
	}, env, svrCtx); err != nil {
```

integrations/integrations.go
```diff
@@ -0,0 +1,231 @@
...
+		return nil, fmt.Errorf("create app db: %v", err)
+	}
+	env.DeferErr("close app db", appdb.Close)
+	blockdb := dbm.NewMemDB()
+	env.DeferErr("close block db", blockdb.Close)
+	txdb := cometdb.NewMemDB()
+	env.DeferErr("close tx db", txdb.Close)
+	mempooldb := dbm.NewMemDB()
+	env.DeferErr("close mempool db", mempooldb.Close)
```
joshklop, line 175: Apologies, I think I was incorrect before. We can use the provided `DBOpener` for the application as used [here](https://github.com/cosmos/cosmos-sdk/blob/929803d6ccbbb25b05dd605ca50f3ea06905e359/server/start.go#L631).

For the Monomer dbs (block db, tx db, mempool db), we need to get creative. Since Monomer is replacing CometBFT from the SDK's perspective, we should use Comet's configuration to figure out what database we should use. The `DefaultDBProvider` as used [here](https://github.com/cosmos/cosmos-sdk/blob/929803d6ccbbb25b05dd605ca50f3ea06905e359/server/start.go#L782) (similar to your usage for the app db I believe) should work for us.

In the tests, we can configure the `DBOpener` (app db) and `DefaultDBProvider` (block db, tx db, mempool db) to use in-memory databases.

Does this hopefully make things clearer? Apologies again for the confusion here.

integrations/integrations.go
```diff
@@ -0,0 +1,231 @@
...
+			},
+			OnCometServeErrCb: func(err error) {
+				svrCtx.Logger.Error("[CometBFT]", "error", err)
+			},
+		},
+	)
+	svrCtx.Logger.Info("Spinning up Monomer node")
+
+	nodeCtx, nodeCtxCancel := context.WithCancel(context.Background())
+	env.Defer(nodeCtxCancel)
```
joshklop, line 221: Let's move this context to the top-level `StartCommandHandler` and pass it down to all functions that require a context. This will also help resolve the context issue mentioned in a comment above.

integrations/integrations.go
```diff
@@ -0,0 +1,231 @@
...
+
+	nodeErr := n.Run(nodeCtx, env)
+	if nodeErr != nil {
```
joshklop, line 224:
```suggestion
	if err := n.Run(nodeCtx, env); err != nil {
```

integrations/integrations.go
```diff
@@ -0,0 +1,231 @@
...
+		return nil, fmt.Errorf("run Monomer node: %v", nodeErr)
+	}
+
+	svrCtx.Logger.Info("Monomer started w/ CometBFT listener on", "address", cometListener.Addr())
+
+	return n, nil
```
joshklop, line 230: no need to return the node if it isn't ever used. Let's just return the error (unless I'm missing a place where the node is used).

integrations/integrations.go
```diff
@@ -0,0 +1,231 @@
...
+func StartCommandHandler(
+	svrCtx *server.Context,
+	clientCtx client.Context, //nolint:gocritic // hugeParam
+	appCreator servertypes.AppCreator,
+	inProcessConsensus bool,
+	opts server.StartCmdOptions,
+) error {
+	// We assume `inProcessConsensus` is true for now, so let's return an error if it's not.
+	if !inProcessConsensus {
+		return fmt.Errorf("in-process consensus must be enabled")
```
joshklop, line 46: very pedantic nit: typical Go style (e.g., [here](https://github.com/uber-go/guide/blob/master/style.md#error-types)) is to use `errors.New` directly for static errors. `fmt.Errorf` uses `errors.New` under the hood I believe.
```suggestion
		return errors.New("in-process consensus must be enabled")
```

integrations/integrations.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	env := environment.New()
+	defer func() {
+		if err := env.Close(); err != nil {
+			svrCtx.Logger.Error("Failed to close environment", "err", err)
```
joshklop, line 52: My bad, I forgot we can handle this error and pass it to the caller instead of punting and printing a log. I have some helper code for this in a separate PR, so we can ignore it for now.

joshklop (commented):
Looks great! Wrapping up the implementation now, just a few more things (mostly that were my fault for unclear explanations).

Once these are resolved, I'll take a closer look at the tests. Great work! I love how you've cut the miscellaneous helper functions to the essentials. Much more maintainable.

integrations/integrations.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	// TODO: Check if is testnet and implement `testnetify` function
+
+	app = appCreator(svrCtx.Logger, db, &fakeTraceWriter{}, svrCtx.Viper)
+	env.Defer(func() {
+		if localErr := app.Close(); localErr != nil {
+			svrCtx.Logger.Error(localErr.Error())
+		}
+	})
```
joshklop, line 113:
```suggestion
	env.DeferErr("close app", app.Close)
```

integrations/integrations.go
```diff
@@ -0,0 +1,231 @@
...
+	return nil
+}
+
+// See https://github.com/cosmos/cosmos-sdk/blob/7fb26685cd68a6c1d199dc270c80f49f2bfe7ace/server/start.go#L624
+func startApp(
+	env *environment.Env,
+	svrCtx *server.Context,
+	appCreator servertypes.AppCreator,
+	opts server.StartCmdOptions,
+) (app servertypes.Application, err error) {
```
joshklop, line 100: Let's avoid named return values unless we need them (I think this is left over the from the Cosmos SDK).
```suggestion
) (servertypes.Application, error) {
```

integrations/integrations.go
```diff
@@ -0,0 +1,231 @@
...
+	db, err := opts.DBOpener(svrCtx.Config.RootDir, server.GetAppDBBackend(svrCtx.Viper))
+	if err != nil {
+		return app, fmt.Errorf("open db: %v", err)
```
joshklop, line 103:
```suggestion
		return nil, fmt.Errorf("open db: %v", err)
```

integrations/integrations.go
phasewalk1: Yes this clears things up. Thank you! I'll make the change

phasewalk1: I noticed that `DefaultDBProvider` returns a `cometbft-db.DB` object, but our `node.New` expects `blockdb` and `mempooldb` to be of type `cosmos-db.DB` ...

integrations/integrations.go
phasewalk1: Is there an `errgroup.New()` function? I only see `WithContext` ... Should I do something like:
```go
if opts.PostSetup != nil {
    g, ctx := errgroup.WithContext(*monomerCtx)
    if err := opts.PostSetup(svrCtx, *clientCtx, ctx, g); err != nil {
        return fmt.Errorf("run post setup: %v", err)
    }
    if err := g.Wait(); err != nil {
        return fmt.Errorf("unexpected error in PostSetup errgroup: %v", err)
    }
}
return nil
```

where `monomerCtx` is the context we open in `StartCommandHandler` and pass down to all other helper functions, eventually using it in the call to `node.Run`

joshklop: > Is there an errgroup.New() function?

It appears there is not, but a [zero errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup#Group) is valid. We can do `var g errgroup.Group` instead.

phasewalk1: > It appears there is not, but a [zero errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup#Group) is valid. We can do var g errgroup.Group instead.

Sounds good. I just made this change and noticed we no longer need `integrations/helpers.go`. Awesome!

integrations/integrations_test.go
```diff
@@ -0,0 +1,190 @@
...
+	}()
+	time.Sleep(1 * time.Second)
+
+	// --- Ensure Monomer has an active CometBFT listener ---
+	if conn, err := net.Dial("tcp", cmtListenAddr); err != nil {
+		svrCtx.Logger.Error("failed to dial CometBFT", "error", err)
+		panic(err)
+	} else {
+		conn.Close()
+	}
```
joshklop, line 80: If we instead do something like `require.True(t, url.Parse(cmtListenAddr).IsReachable(ctx))`, we can avoid the sleep. Correct?

integrations/integrations_test.go
```diff
@@ -0,0 +1,190 @@
...
+	svrCtx.Logger.Info("CometBFT client created", "bftClient", bftClient)
+
+	txBytes := testapp.ToTx(t, "userTxKey", "userTxValue")
+	bftTx := bfttypes.Tx(txBytes)
+
+	putTx, err := bftClient.BroadcastTxAsync(ctx, txBytes)
+	require.NoError(t, err, "could not broadcast tx")
+	require.Equal(t, abcitypes.CodeTypeOK, putTx.Code, "put.Code is not OK")
+	require.EqualValues(t, bftTx.Hash(), putTx.Hash, "put.Hash is not equal to bftTx.Hash")
+	svrCtx.Logger.Info("Monomer Tx broadcasted successfully", "txHash", putTx.Hash)
```
joshklop, line 96: prefer `t.Info` to control log output in tests.

integrations/integrations_test.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// Wrapper around `testapp.App` to satisfy the `ABCI` and `servertypes.Application` interfaces
+type WrappedTestApp struct {
+	*testapp.App
```
joshklop, line 101: Does a named field here `app *testapp.App` work?

integrations/integrations_test.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// ---- `ABCI` interface ----
```
joshklop, line 104: Rather than including these comments, it would be preferable to do something like:

```golang
// (not sure if `servertypes.ABCI` is correct, it's just an example
var _ servertypes.ABCI = (*WrappedTestApp)(nil)
```
That way interface compliance is checked by the compiler. See [here](https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance) for details.

integrations/integrations_test.go
phasewalk1: I see `t.Log` and `t.Errorf` but no `t.Info` method

integrations/integrations_test.go
phasewalk1: I tried but it introduces an error: `cannot use WrappedTestApp as serevertypes.Application`

integrations/integrations_test.go
joshklop: My bad, meant `t.Log`.

integrations/integrations_test.go
phasewalk1: Yes good point. This will allow us to avoid the sleep.

integrations/integrations_test.go
```diff
@@ -0,0 +1,190 @@
...
+	opts := server.StartCmdOptions{
+		DBOpener: func(rootDir string, backendType dbm.BackendType) (dbm.DB, error) {
+			return dbm.NewMemDB(), nil
+		},
+	}
+
+	cmtListenAddr := svrCtx.Config.RPC.ListenAddress
+	cmtListenAddr = strings.TrimPrefix(cmtListenAddr, "tcp://")
+
+	go func() {
```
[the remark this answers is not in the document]
phasewalk1: Hmm. I tried to follow the same pattern we use `node/http.go` but the test seems to hang. Maybe I'm doing something wrong, but this is what I had:
```go
var wg conc.WaitGroup
defer wg.Wait()
wg.Go(func() {
    err := StartCommandHandler(svrCtx, clientCtx, mockAppCreator, inProcessConsensus, opts)
    require.NoError(t, err)
})
```
Do we need to use an error channel / `select` block?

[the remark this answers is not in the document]
joshklop: I think this is expected. Here's what I think is happening:

1. `StartCommandHandler` runs in the goroutine
2. It sets up all the services and then blocks on the signal channel
3. No signal is ever sent
4. We wait forever in the test for `StartCommandHandler` to finish

I think we will need an ugly hack to make this work. Instead of defining the signal channel in `StartCommandHandler`, we can make it a global variable.

```golang
var sigChan = make(chan signal.Signal, 1) // not sure if I got the naming right, but you get the idea
```

Now, we can send a signal on this channel at the end of the test to unblock `StartCommandHandler`. Does this make sense?

integrations/integrations_test.go
joshklop: Ah I see. We have the `Query` method defined on the testapp. Disregard

integrations/integrations.go
```diff
@@ -0,0 +1,246 @@
...
+	if err != nil {
+		return fmt.Errorf("create CometBFT listener: %v", err)
+	}
+	env.DeferErr("close comet listener", cometListener.Close)
+
+	appdb, err := opts.DBOpener(svrCtx.Config.RootDir, server.GetAppDBBackend(svrCtx.Viper))
+	if err != nil {
+		return fmt.Errorf("open app db: %v", err)
+	}
+	env.DeferErr("close app db", appdb.Close)
```
joshklop, line 176: unused? We build the app up above, no need for the db here.

integrations/integrations.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+	env.DeferErr("close tx db", txdb.Close)
+
+	mempooldb, err := dbm.NewDB("mempool", dbm.BackendType(svrCtx.Config.DBBackend), svrCtx.Config.RootDir)
+	if err != nil {
+		return fmt.Errorf("create mempool db: %v", err)
+	}
+	env.DeferErr("close mempool db", mempooldb.Close)
+
+	monomerGenesisPath := viper.GetString(monomerGenesisPathFlag)
```
joshklop, line 196: I just realized we can use `svrCtx.Config.GenesisFile()` instead of adding an environment variable.

integrations/integrations_test.go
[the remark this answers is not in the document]
phasewalk1: Yes this makes sense, thank you. Here's what I have so far, but the test is still hanging ...

```go
// integrations.go

var sigCh = make(chan os.Signal, 1)

// ...

func StartCommandHandler(...) {
    // ...
    signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
    <-sigCh
}
```

```go
func TestStartCommandHandler(...) {
    // ...
    var wg conc.WaitGroup
    defer wg.Wait()
    wg.Go(func() {
        err := StartCommandHandler(svrCtx, clientCtx, mockAppCreator, inProcessConsensus, opts)
        require.NoError(t, err)
    })
    // ...
    sigCh <- syscall.SIGINT
}
```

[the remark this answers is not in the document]
joshklop: Where is it hanging?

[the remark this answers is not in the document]
phasewalk1: It logs that the Tx broadcasted successfully, so it appears to hang while waiting for a signal on the channel

PR #44, opened 2024-06-18
Replace db backend

joshklop:
## Summary by CodeRabbit

- **New Features**
	- Introduced a `DB` interface for enhanced interaction with blockchain data.
	- Transitioned to a Pebble database for improved performance and capabilities.
	- Updated block management methods to utilize the new interface.

- **Bug Fixes**
	- Enhanced error handling in block retrieval and appending operations.

- **Documentation**
	- Updated comments and structuring to improve clarity regarding database interactions and block management processes.

- **Chores**
	- Updated dependencies in the module to include `pebble` and `cbor`.

- **Tests**
	- Added comprehensive test cases for the new database functionalities and improved error handling in existing tests.


monomer.go
```diff
@@ -124,11 +123,9 @@ func (b *Block) ToEth() (*ethtypes.Block, error) {
...
 func (b *Block) ToCometLikeBlock() *bfttypes.Block {
 	return &bfttypes.Block{
-		Header: bfttypes.Header{
-			ChainID: b.Header.ChainID.String(),
-			Time:    time.Unix(int64(b.Header.Time), 0),
-			Height:  b.Header.Height,
-			AppHash: bftbytes.HexBytes(b.Header.AppHash),
+		Header: *b.Header.ToComet(),
+		Data: bfttypes.Data{
+			Txs: b.Txs,
```
joshklop, line 128: reminder to myself to move this to the next commit

monomerdb/localdb/localdb.go
```diff
@@ -0,0 +1,464 @@
...
+			UpperBound: nextBucket,
+		})
+		if err != nil {
+			return fmt.Errorf("new bucketHeaderByHeight iterator: %v", err)
+		}
+		defer func() {
+			err = wrapCloseErr(err, iter)
+		}()
+		header := new(monomer.Header)
+		for iter.First(); iter.Valid(); iter.Next() {
```
natebeauregard, line 122: If `iter.Valid()` returns false then we'll exit this for loop. Are there any cases where we may hit an invalid `iter` due to a malformed entry or DB corruption but we still want to delete subsequent headers (or transactions down below)?

monomerdb/localdb/localdb.go
```diff
@@ -0,0 +1,464 @@
...
+		txs = append(txs, tx)
+	}
+	return txs, nil
+}
+
+type getter interface {
+	Get([]byte) ([]byte, io.Closer, error)
+}
+
+func headerByHeight(g getter, heightBytes []byte) (_ *monomer.Header, err error) {
```
natebeauregard, line 356: For my own learning about go styling conventions, when should you use named return values instead of nameless return values?

monomerdb/localdb/localdb.go
```diff
@@ -0,0 +1,464 @@
...
+	return nil
+}
+
+func get(g getter, key []byte) (_ []byte, _ io.Closer, err error) {
+	value, closer, err := g.Get(key)
+	if err != nil {
+		if errors.Is(err, pebble.ErrNotFound) {
+			return nil, nil, monomerdb.ErrNotFound
+		}
+		return nil, nil, fmt.Errorf("%v", err) // Obfuscate the error.
```
natebeauregard, line 444: Is `fmt.Errorf("%v", err)` the same as just returning `err` directly?

monomerdb/localdb/localdb.go
```diff
@@ -0,0 +1,464 @@
...
+	return db.updateIndexed(func(b *pebble.Batch) (err error) {
+		unsafeHeightBytesValue, closer, err := get(b, bucketHeightByHash.Key(unsafe.Bytes()))
+		if err != nil {
+			return fmt.Errorf("get height by hash %s: %w", unsafe, err)
+		}
+		defer func() {
+			err = wrapCloseErr(err, closer)
+		}()
+		unsafeHeight := endian.Uint64(unsafeHeightBytesValue)
+		firstHeightBytesToDelete := marshalUint64(unsafeHeight + 1)
```
natebeauregard, line 101: curious - why do we rollback to the unsafe height here instead of the safe height?

comet/comet.go
```diff
@@ -70,7 +70,7 @@ func (s *ABCI) Info(ctx *jsonrpctypes.Context) (*rpctypes.ResultABCIInfo, error)
 }
 
 type HeadBlocker interface {
-	HeadBlock() *monomer.Block
+	HeadHeader() (*monomer.Header, error)
```
natebeauregard, line 73: Should the `HeadBlocker` interface be renamed now that it only returns the header?

integrations/integrations.go
```diff
@@ -163,11 +166,21 @@ func startMonomerNode(
...
+	if backendType := dbm.BackendType(svrCtx.Config.DBBackend); backendType == dbm.MemDBBackend {
+		blockPebbleDB, err = pebble.Open("", &pebble.Options{
+			FS: vfs.NewMem(),
+		})
+	} else {
+		if backendType != dbm.PebbleDBBackend {
+			svrCtx.Logger.Info("Overriding provided db backend for the blockstore", "provided", backendType, "using", dbm.PebbleDBBackend)
+		}
+		blockPebbleDB, err = pebble.Open("blockstore", nil)
+	}
```
natebeauregard, line 179: Should we still use `svrCtx.Config.RootDir` as the pebble directory name whenever the DB is opened?

natebeauregard (commented):
Really cool PR! Everything looks good, I left a few questions to ensure I was following everything correctly and a few small suggestions.

monomerdb/localdb/localdb.go
joshklop: > Are there any cases where we may hit an invalid iter due to a malformed entry or DB corruption

This should never ever ever happen. Very very bad. All bets are off when the db is corrupted.

monomerdb/localdb/localdb.go
joshklop: Slightly different. Returning the err directly would leak the error type information to the caller (so they can do `errors.Is`, etc.). Technically, this would expand our API to include pebble errors. We wrap errors with `%v` to limit the API surface.

monomerdb/localdb/localdb.go
joshklop: There are always three heights, unsafe, safe, and finalized. 

unsafe >= safe >= finalized

If we roll back to safe, we might delete unsafe blocks.

integrations/integrations.go
joshklop: Ahhh good point. I think you're right. Will investigate.

monomerdb/localdb/buckets.go
```diff
@@ -0,0 +1,24 @@
+package localdb
+
+import (
+	"bytes"
+)
+
+type dbBucket byte
+
+const (
+	bucketHeaderByHeight dbBucket = iota + 1
```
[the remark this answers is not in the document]
joshklop: `iota` is the default value. If there's a bug where we try to create a key using an uninitialized `dbBucket` var, we at least won't cause db corruption:

```go
var b dbBucket
b.Key() // Key will not exist in the db.
```

https://github.com/uber-go/guide/blob/master/style.md#start-enums-at-one

monomerdb/localdb/localdb.go
joshklop: Named return values can be tricky. I only use them for wrapping errors in `defer`s. I prefer to keep all other return values declared inside the function itself.

From the go spec on [declarations](https://go.dev/ref/spec#Declarations_and_scope)
> The blank identifier may be used like any other identifier in a declaration, but it does not introduce a binding and thus is not declared.

Also see the spec for the [blank identifier](https://go.dev/ref/spec#Blank_identifier)

joshklop: This is just my habit though. I haven't seen good alternatives yet, but happy to change if someone shows me something better.

monomerdb/localdb/localdb.go
```diff
@@ -0,0 +1,464 @@
...
+		}
+
+		if err := updateLabels(b, unsafe, safe, finalized); err != nil {
+			return fmt.Errorf("update labels: %v", err)
+		}
+		return nil
+	})
+}
+
+func (db *DB) HeadHeader() (*monomer.Header, error) {
```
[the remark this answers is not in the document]
joshklop: This actually was an important change: we were leaking an iterator (not closing) in `localdb.DB.Rollback` because we were reassigning it before closing it. Closing the db caught the issue!

monomer.go
joshklop: ultimately decided against this; would take an unnecessarily long time to do this. The real solution is to break up such large PRs :)

natebeauregard (approved):
Changes LGTM! Feel free to merge after fixing up the commits

monomerdb/localdb/localdb.go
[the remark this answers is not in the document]
joshklop: woops, meant to leave the above response on a separate comment.

I ended up deciding against renaming `Head`->`Latest` for now, since `latest` can be a synonym for `unsafe` in Ethereum-speak (especially in the JSON-RPC API). To avoid confusion between the head block in the db and the `unsafe` block, I'll leave it as it is for now.

PR #54, opened 2024-07-09
Stop builder_test when subscription channel is closed

natebeauregard:
Makes the builder test stop when the subscription channel is closed. Wraps the event bus logic in a `select` statement to ensure that the subscription channel is still open before reading the event.

## Summary by CodeRabbit

- **Tests**
  - Improved event handling in test cases for better robustness and reliability.

### DIFF
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -161,7 +161,6 @@ func TestBuild(t *testing.T) {
 			require.Equal(t, wantBlock, gotBlock)
 
 			// Tx store and event bus.
-			eventChan := subscription.Out()
 			for i, tx := range wantBlock.Txs {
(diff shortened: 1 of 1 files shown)


NiloCK (approved):
This LGTM, but a minor question/suggestion:

Is the `eventChan` variable useful here? Seems like using `subscription.Out()` directly is more direct and readable. The `select` statement would have two reads from the subscription struct.

natebeauregard:
> This LGTM, but a minor question/suggestion:
> 
> Is the `eventChan` variable useful here? Seems like using `subscription.Out()` directly is more direct and readable. The `select` statement would have two reads from the subscription struct.

Good point, I'll remove the `eventChan` var and inline `subscription.Out()` directly.

PR #58, opened 2024-07-16
Add prometheus server and basic eth/engine rpc metrics

natebeauregard:
### Change Overview
Adds a prometheus server to the monomer node that's configured through the [CometBFT Instrumentation config](https://docs.cometbft.com/v0.37/core/configuration) in the cosmos SDK app's `config.toml` file. These changes will allow for easy Prometheus/Grafana integrations. Additionally, adds basic eth and engine RPC metrics:
- Count and duration (histogram buckets) of the following `eth` RPC calls:
  - `eth_getChainId`
  - `eth_getBlockByHash`
  - `eth_getBlockByNumber`
- Count and duration (histogram buckets) of the following `engine` RPC calls:
  - `engine_forkchoiceUpdatedV3`
  - `engine_getPayloadV3`
  - `engine_newPayloadV3`

[Sample prometheus server response](https://gist.github.com/natebeauregard/684bc43ac98a00f8571ba944f4732000)

### Local configuration
To test metrics locally, setup up a monomer node with prometheus enabled in the CometBFT `config.toml` file (or just inject directly into the server context). Here is an example instrumentation config with prometheus enabled:
```
[instrumentation]
prometheus = true
prometheus_listen_addr = ":26660"
max_open_connections = 1
namespace = "monomer"
```

Then, either query the prometheus metrics directly at http://localhost:26660/metrics or start up a local prometheus server by following the prometheus docs for [installation](https://prometheus.io/docs/prometheus/latest/installation/) and using this scrape config in `prometheus.yml`:
```
scrape_configs:
  - job_name: "prometheus"
    static_configs:
      - targets: ["localhost:9090"]

  - job_name: "monomer"
    static_configs:
      - targets: ["localhost:26660"]
```

## Summary by CodeRabbit

- **New Features**
    - Introduced instrumentation configuration for better monitoring and diagnostics.

- **Dependency Changes**
    - Added direct dependencies:
        - `cosmossdk.io/store v1.1.0`
        - `github.com/prometheus/client_golang v1.19.1`
        - `github.com/spf13/viper v1.18.2`
        - `golang.org/x/sync v0.7.0`

- **Tests**
    - Enhanced test suite with HTTP request handling and instrumentation settings.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log"
 	"sync"
+	"time"
 
 	abci "github.com/cometbft/cometbft/abci/types"
 	"github.com/ethereum-optimism/optimism/op-service/eth"
@@ -29,6 +30,7 @@ type EngineAPI struct {
 	txValidator              TxValidator
 	blockStore               BlockStore
 	currentPayloadAttributes *monomer.PayloadAttributes
+	metrics                  Metrics
 	lock                     sync.RWMutex
 }
 
@@ -40,11 +42,13 @@ func NewEngineAPI(
 	b *builder.Builder,
 	txValidator TxValidator,
 	blockStore BlockStore,
+	metrics Metrics,
 ) *EngineAPI {
 	return &EngineAPI{
 		txValidator: txValidator,
 		blockStore:  blockStore,
 		builder:     b,
+		metrics:     metrics,
 	}
 }
 
@@ -73,6 +77,7 @@ func (e *EngineAPI) ForkchoiceUpdatedV3(
 ) (*eth.ForkchoiceUpdatedResult, error) {
 	e.lock.Lock()
 	defer e.lock.Unlock()
+	defer e.metrics.RecordRPCMethodCall(ForkchoiceUpdatedV3MethodName, time.Now())
 
 	// OP spec:
 	//   - headBlockHash: block hash of the head of the canonical chain. Labeled "unsafe" in user JSON-RPC.
@@ -233,6 +238,7 @@ func (e *EngineAPI) GetPayloadV2(ctx context.Context, payloadID engine.PayloadID
 func (e *EngineAPI) GetPayloadV3(ctx context.Context, payloadID engine.PayloadID) (*eth.ExecutionPayloadEnvelope, error) {
 	e.lock.RLock()
 	defer e.lock.RUnlock()
+	defer e.metrics.RecordRPCMethodCall(GetPayloadV3MethodName, time.Now())
 
 	if e.currentPayloadAttributes == nil {
 		return nil, engine.InvalidParams.With(errors.New("payload not found"))
@@ -301,6 +307,7 @@ func (e *EngineAPI) NewPayloadV2(payload eth.ExecutionPayload) (*eth.PayloadStat
 func (e *EngineAPI) NewPayloadV3(payload eth.ExecutionPayload) (*eth.PayloadStatusV1, error) { //nolint:gocritic
 	e.lock.Lock()
 	defer e.lock.Unlock()
+	defer e.metrics.RecordRPCMethodCall(NewPayloadV3MethodName, time.Now())
 
 	if e.blockStore.BlockByHash(payload.BlockHash) == nil {
 		return &eth.PayloadStatusV1{
--- a/engine/metrics.go
+++ b/engine/metrics.go
@@ -13,7 +13,7 @@ const (
 )
 
 var (
-	RPCMethodDurationBucketsMicroseconds = []float64{1, 10, 50, 100, 500, 1000, 5000, 10000, 100000}
+	RPCMethodDurationBucketsMicroseconds = []float64{1, 10, 50, 100, 500, 1000}
 )
 
 // Metrics contains metrics collected from the engine package.
--- a/engine/metrics.go
+++ b/engine/metrics.go
@@ -12,9 +12,7 @@ const (
 	NewPayloadV3MethodName        = "newPayloadV3"
 )
 
-var (
-	RPCMethodDurationBucketsMicroseconds = []float64{1, 10, 50, 100, 500, 1000}
-)
+var RPCMethodDurationBucketsMicroseconds = []float64{1, 10, 50, 100, 500, 1000}
 
 // Metrics contains metrics collected from the engine package.
 type Metrics interface {
--- a/engine/metrics.go
+++ b/engine/metrics.go
@@ -8,8 +8,8 @@ const (
 	MetricsSubsystem = "engine"
 
 	ForkchoiceUpdatedV3MethodName = "forkchoiceUpdatedV3"
-	GetPayloadV3MethodName        = "getPayload"
-	NewPayloadV3MethodName        = "newPayload"
+	GetPayloadV3MethodName        = "getPayloadV3"
+	NewPayloadV3MethodName        = "newPayloadV3"
 )
 
 var (
--- a/engine/metrics.go
+++ b/engine/metrics.go
@@ -2,7 +2,6 @@ package engine
 
 import (
 	rpcmetrics "github.com/polymerdao/monomer/metrics"
-	"time"
 )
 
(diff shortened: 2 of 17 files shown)


NiloCK:
> To test metrics locally, setup up a monomer node with prometheus enabled in the CometBFT config.toml file (or just inject directly into the server context). Here is an example instrumentation config with prometheus enabled:

Where should that `config.toml` file be? Is there one in repo now?

natebeauregard:
> > To test metrics locally, setup up a monomer node with prometheus enabled in the CometBFT config.toml file (or just inject directly into the server context). Here is an example instrumentation config with prometheus enabled:
> 
> Where should that `config.toml` file be? Is there one in repo now?

@NiloCK The `config.toml` file will exist in the repo of the Cosmos SDK chain that's using monomer for its consensus. That file is [parsed at the appchain level](https://github.com/cosmos/cosmos-sdk/blob/afddef3/server/util.go#L132) and is then [passed through to monomer](https://github.com/polymerdao/monomer/blob/3d13d20/integrations/integrations.go#L38) in the `StartCommandHandler` as `svrCtx.Config`. For testing, we can just set the `svrCtx.Config` values directly in the monomer `StartCommandHandler` since those values will already be set before entering the monomer repo.

Existing example in `integrations_test.go`:
https://github.com/polymerdao/monomer/blob/3d13d206de58eaa166812df1cee798575f03fd39/integrations/integrations_test.go#L50-L54

engine/engine.go
```diff
@@ -73,6 +77,7 @@ func (e *EngineAPI) ForkchoiceUpdatedV3(
 ) (*eth.ForkchoiceUpdatedResult, error) {
 	e.lock.Lock()
 	defer e.lock.Unlock()
+	defer e.metrics.RecordRPCMethodCall(ForkchoiceUpdatedV3MethodName, time.Now())
```
NiloCK, line 80: For my understanding:

the `time.Now()` gets called "as this is read", so that start time of the function is available to the metrics recording function.

engine/metrics.go
```diff
@@ -0,0 +1,43 @@
...
+func NewMetrics(namespace string) Metrics {
+	return &metrics{
+		rpcmetrics.NewRPCMetrics(
+			namespace,
+			MetricsSubsystem,
+			"Duration of each engine RPC method call in microseconds",
+			RPCMethodDurationBucketsMicroseconds,
+		),
+	}
+}
```
NiloCK, line 35:
```suggestion
func NewMetrics(namespace string) Metrics {
	rpcm := rpcmetrics.NewRPCMetrics(
		namespace,
		MetricsSubsystem,
		"Duration of each engine RPC method call in microseconds",
		RPCMethodDurationBucketsMicroseconds,
	)
	return &rpcm
}
```

Seems to work. Is there a specific need for the indirection here?

engine/metrics.go
```diff
@@ -0,0 +1,43 @@
...
+
+type noopMetrics struct {
+	rpcmetrics.RPCNoopMetrics
+}
+
+func NewNoopMetrics() Metrics {
+	return &noopMetrics{}
+}
```
NiloCK, line 43: Similar:

```suggestion
func NewNoopMetrics() Metrics {
	return &rpcmetrics.RPCNoopMetrics{}
}
```

eth/metrics.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+}
+
+type noopMetrics struct {
+	rpcmetrics.RPCNoopMetrics
+}
+
+func NewNoopMetrics() Metrics {
+	return &noopMetrics{}
+}
```
NiloCK, line 43: Comments in `engine/metrics.go` reapply here I think.

metrics/rpc_metrics.go
```diff
@@ -0,0 +1,40 @@
...
+
+	stdprometheus "github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+type Metrics interface {
+	RecordRPCMethodCall(method string, start time.Time)
+}
+
+type RPCMetrics struct {
```
NiloCK, line 14: Huge nit but this type is applied against both eth json `RPC` methods and engine `API` methods.

Slightly annoying, but I think not worth a lot of consideration.

node/node_test.go
```diff
@@ -47,6 +50,12 @@ func TestRun(t *testing.T) {
 		blockdb,
 		mempooldb,
 		txdb,
+		&config.InstrumentationConfig{
+			Prometheus:           true,
+			PrometheusListenAddr: "127.0.0.1:26660",
```
NiloCK, line 55: Maybe: pull this to an in-test const.

NiloCK (approved):
This LGTM! Some small suggestions. Might be nice to turn these on in the e2e test and assert some expected behaviours against the recorded metrics.

metrics/rpc_metrics.go
```diff
@@ -0,0 +1,40 @@
+package metrics
+
+import (
+	"time"
+
+	stdprometheus "github.com/prometheus/client_golang/prometheus"
```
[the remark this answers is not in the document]
natebeauregard: Good catch, I ran `go mod tidy` and the prometheus dependency and a few others were added in for the top level `go.mod` file

engine/engine.go
natebeauregard: Yep, the start time whenever the RPC method was originally called will be evaluated immediately and passed into `RecordRPCMethodCall` and will then be subtracted from the current time whenever the RPC method completes.

engine/metrics.go
natebeauregard: Yeah I kept going back and forth on this. If we eventually want to add other metrics to the eth or engine packages outisde of the shared RPC metrics then we'd have to implement them individually in the `eth/metrics.go` or `engine/metrics.go` file and could then easily add them here by adding a line to the metrics struct. However, if we only anticipate having the shared RPC metrics then your suggested change would be fine. Can you think of any other metrics that we may want to eventually add for either of these packages outside of the RPC method call/duration metrics?

cc @joshklop

engine/metrics.go
natebeauregard: Same response as [above](https://github.com/polymerdao/monomer/pull/58#discussion_r1681637927)

eth/metrics.go
natebeauregard: Same response as [above](https://github.com/polymerdao/monomer/pull/58#discussion_r1681637927)

node/prometheus.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (n *Node) registerMetrics() (eth.Metrics, engine.Metrics) {
+	if n.prometheusCfg.IsPrometheusEnabled() {
+		namespace := n.prometheusCfg.Namespace
+		return eth.NewMetrics(namespace),
+			engine.NewMetrics(namespace)
+	}
+	return eth.NewNoopMetrics(),
+		engine.NewNoopMetrics()
+}
```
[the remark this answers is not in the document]
natebeauregard: We could add support for that but after looking at how CometBFT and OP Stack metrics work it seems to be an all or nothing approach instead of an individual configuration for enabling/disabling metrics for individual packages. To avoid needing to update the prometheus configuration I'm going to opt to leave this as is for now but we can always refactor this to support enabling/disabling metrics for individual packages if that turns out to be a common use case

metrics/rpc_metrics.go
natebeauregard: Good point, I think it's probably worth renaming. I guess that all RPC methods are also technically [API methods](https://ethereum.org/en/developers/docs/apis/json-rpc/) , do you think that renaming this to `APIMetrics` would be more descriptive?

NiloCK: Could do. Could leave as is. Up to you :)

natebeauregard: On second thought I'll opt to just leave it as is since the `engine` methods are also RPC calls as well

PR #59, opened 2024-07-16
`testapp`: Enable Default AnteHandler

phasewalk1:
## Overview
Makes our `testapp.App` more realistic by enabling the default `AnteHandler` in the testapp's `tx` module. 

> [!IMPORTANT]
> `AnteHandlers` perform basic validity checks on a transaction, such that it could be thrown out of the mempool. They are called on both `CheckTx` and `DeliverTx`, and check things like:
> - The feepayer address exists and has enough funds to pay the gas fee
> - The transaction signature is valid and the signer has the correct authority

Tagging: @natebeauregard @joshklop for review

## Summary by CodeRabbit

- **New Features**
  - Introduced new constants across test files for improved chain ID management.
  - Enhanced transaction handling logic in tests to reflect real-world scenarios.
  - Added private key management features for secure operations.
  - Implemented dynamic genesis data handling for increased application flexibility.
  - Streamlined application state initialization for more accurate test environments.

- **Bug Fixes**
  - Improved robustness and accuracy of end-to-end tests by refining transaction creation and state initialization.

- **Chores**
  - Updated genesis file to include essential account and balance details for better initialization.


comet/comet_test.go
```diff
@@ -116,13 +125,24 @@ func TestStatus(t *testing.T) {
...
+		AppStateBytes: func() []byte {
+			appStateBytes, err := json.Marshal(testapp.MakeGenesisAppState(t, app))
+			require.NoError(t, err)
+			return appStateBytes
+		}(),
+	})
+	require.NoError(t, err)
+
+	ctx := app.GetContext(true)
+	sk, _, acc := app.TestAccount(ctx)
```
phasewalk1, line 140: This is a common pattern across every test that needs to sign transactions with `testapp.ToTx`. We need to call `InitChain` before we fetch the application context, and then we get our pre-funded test account with `app.TestAccount`.

It'd probably be good to encapsulate this setup logic in a helper function (perhaps inside of `testapp.NewTest` itself), so we can avoid repeating it across each test.

comet/comet_test.go
```diff
@@ -116,13 +125,24 @@ func TestStatus(t *testing.T) {
 }
 
 func TestBroadcastTx(t *testing.T) {
-	chainID := "0"
-	app := testapp.NewTest(t, chainID)
+	app := testapp.NewTest(t, chainID.String())
```
phasewalk1, line 128: This is a common pattern across every test that needs to sign transactions with `testapp.ToTx`. We need to call `InitChain` before we fetch the application context, and then we get our pre-funded test account with `app.TestAccount`.

It'd probably be good to encapsulate this setup logic in a helper function (perhaps inside of `testapp.NewTest` itself), so we can avoid repeating it across each test.

comet/comet_test.go
phasewalk1, line 140: This is a common pattern across every test that needs to sign transactions with testapp.ToTx. We need to call InitChain before we fetch the application context, and then we get our pre-funded test account with app.TestAccount.

It'd probably be good to encapsulate this setup logic in a helper function (perhaps inside of testapp.NewTest itself), so we can avoid repeating it across each test.

e2e/stack_test.go
```diff
@@ -158,7 +162,22 @@ func TestE2E(t *testing.T) {
 	client, err := bftclient.New(monomerCometURL.String(), monomerCometURL.String())
 	require.NoError(t, err, "create Comet client")
 
-	txBytes := testapp.ToTx(t, "userTxKey", "userTxValue")
+	app := testapp.NewTest(t, chainID.String())
```
phasewalk1, line 165: This is a common pattern across every test that needs to sign transactions with testapp.ToTx. We need to call InitChain before we fetch the application context, and then we get our pre-funded test account with app.TestAccount.

It'd probably be good to encapsulate this setup logic in a helper function (perhaps inside of testapp.NewTest itself), so we can avoid repeating it across each test.

comet/comet_test.go
phasewalk1, line 140: This is a common pattern across every test that needs to sign transactions with testapp.ToTx. We need to call InitChain before we fetch the application context, and then we get our pre-funded test account with app.TestAccount.

It'd probably be good to encapsulate this setup logic in a helper function (perhaps inside of testapp.NewTest itself), so we can avoid repeating it across each test.

joshklop: That sounds great! Should make the test much easier to review. The diff is a bit scary atm :)

phasewalk1: Agreed, that was my intention with this comment. I realized the diff was a bit scary 😆. Will re-ping when RFR.

phasewalk1: @joshklop I thought for a while about the best way to simplify the changes here, and settled on refactoring `testapp.NewTest` to encapsulate this repeated setup logic for us. However, there are some tests that call `genesis.Commit` and some that don't -- and there are some tests that need access to a `blockStore` and `ethstatedb`, while others don't.

My idea for a solution here is to create a struct like this:
```go
// TestApp wraps a boilerplate Cosmos SDK application with auxiliary data structures for testing.
type TestApp struct {
	*App
	Ctx        sdk.Context
	ChainID    monomer.ChainID
	Genesis    *genesis.Genesis
	BlockStore *localdb.DB
	EthStateDB state.Database
	PrivKey    *secp256k1.PrivKey
	Account    sdk.AccountI
}
```
and have `NewTest` call `InitChain` for us (and optionally `genesis.Commit`) then return this structure. That way the application is already initialized for us in every test, and each test has access to these auxiliary components if they need them. I also made a few other changes to helper functions to shrink the footprint of this diff, though please feel free to lmk if there's more I can do to make review easier.

phasewalk1:
Currently rebasing on top of main and running into a weird issue where `e2e/stack_test.go` hangs before the [_"Monomer can sync"_ stage](https://github.com/polymerdao/monomer/blob/35910beffd4155644df656cfe6ea63ebc7526086/e2e/stack_test.go#L185).

joshklop:
With #267, this will no longer be necessary. Thanks for the hard work: this effort informed future feature and test development.

PR #61, opened 2024-07-18
removing Transaction field from the PayloadAttributes structure

AnkushinDaniil:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling during payload validation to enhance reliability.

- **Refactor**
  - Updated methods for better handling of payload attributes and transaction adaptation.
  - Removed outdated fields in payload structure, streamlining transaction processing.

These changes result in a more reliable and efficient transaction validation experience for users.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -208,6 +208,7 @@ func (e *EngineAPI) ForkchoiceUpdatedV3(
 		ParentBeaconBlockRoot: pa.ParentBeaconBlockRoot,
 		ParentHash:            fcs.HeadBlockHash,
 		Height:                e.blockStore.HeadBlock().Header.Height + 1,
+		CosmosTxs:             cosmosTxs,
 	}
 
 	// Engine API spec:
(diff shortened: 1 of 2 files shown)


AnkushinDaniil:
@joshklop review request

joshklop (changes requested):
Thinking about this again, I feel we can simplify things a bit more. Let's remove the `Transactions` field instead of the `CosmosTxs` field. Existing references to `Transactions` can be replaced with `CosmosTxs`.

If you have different thoughts, let me know. My quick implementation locally seems to work well.

engine/engine.go
```diff
@@ -208,8 +208,6 @@ func (e *EngineAPI) ForkchoiceUpdatedV3(
 		ParentBeaconBlockRoot: pa.ParentBeaconBlockRoot,
 		ParentHash:            fcs.HeadBlockHash,
 		Height:                e.blockStore.HeadBlock().Header.Height + 1,
-		Transactions:          pa.Transactions,
-		CosmosTxs:             cosmosTxs,
```
joshklop, line 212: We should keep the `cosmosTxs` field.

AnkushinDaniil: The problem was fixed by re-running `make setup-e2e'.

joshklop (approved):

PR #62, opened 2024-07-19
E2E Test User funding

NiloCK:
PR adds functionality to fund L1 user accounts (both node operator & end-user). Includes a regression check against unfunded L1 operator keys.

## Summary by CodeRabbit

- **New Features**
	- Implemented a balance verification step for operators before executing further actions, ensuring sufficient funds are available.
	- Added support for managing multiple Layer 1 user accounts, enhancing the testing and deployment capabilities for Layer 1 contracts.
  
- **Bug Fixes**
	- Improved error handling for balance retrieval and verification to provide clearer feedback to users.

### DIFF
--- a/e2e/op_stack.go
+++ b/e2e/op_stack.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
+	"errors"
 	"time"
 
 	"github.com/ethereum-optimism/optimism/op-batcher/batcher"
--- a/e2e/op_stack.go
+++ b/e2e/op_stack.go
@@ -85,9 +85,8 @@ func (op *OPStack) Run(ctx context.Context, env *environment.Env) error {
 	balance, err := l1.BalanceAt(ctx, crypto.PubkeyToAddress(op.privKey.PublicKey), nil)
 	if err != nil {
 		return fmt.Errorf("get balance: %v", err)
-	}
-	if balance.Cmp(big.NewInt(0)) == 0 {
-		return fmt.Errorf("stack operator balance is 0")
+	} else if balance.Cmp(big.NewInt(0)) == 0 {
+		return errors.New("stack operator balance is 0")
 	}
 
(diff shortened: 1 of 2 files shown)


e2e/op_stack.go
```diff
@@ -79,6 +81,15 @@ func (op *OPStack) Run(ctx context.Context, env *environment.Env) error {
 		return err
 	}
 
+	// check balance of privKey - operator must hold an L1 balance to post state roots & batches
+	balance, err := l1.BalanceAt(ctx, crypto.PubkeyToAddress(op.privKey.PublicKey), nil)
+	if err != nil {
+		return fmt.Errorf("get balance: %v", err)
+	}
+	if balance.Cmp(big.NewInt(0)) == 0 {
+		return fmt.Errorf("stack operator balance is 0")
```
joshklop, line 90:
```suggestion
	} else if balance.Cmp(big.NewInt(0)) == 0 {
		return errors.New("stack operator balance is 0")
```

e2e/stack.go
```diff
@@ -36,6 +39,12 @@ type Stack struct {
 	eventListener    EventListener
 	l1BlockTime      uint64
 	prometheusCfg    *config.InstrumentationConfig
+	L1Users          []L1User
+}
+
+type L1User struct {
+	Address    common.Address
+	PrivateKey *ecdsa.PrivateKey
```
joshklop, line 47: Is this abstraction necessary? I think it would be better to add something like this when we have the need for multiple L1 users (in that case, it would probably best to keep it private or off of the `Stack` struct entirely).

joshklop (approved):

e2e/op_stack.go
```diff
@@ -5,6 +5,7 @@ import (
 	"crypto/ecdsa"
 	"fmt"
 	"math"
+	"math/big"
```
NiloCK, line 8:
```suggestion
	"math/big"
	"errors"
```

PR #67, opened 2024-07-24
Add MakeBlock constructor

joshklop:
Allows us to ensure blocks are valid (up to stateless validation).

These changes surfaced a bug in the genesis package where the AppHash was incorrectly attached to the genesis header (where it should be attached to the subsequent header). This is fixed in this commit.


joshklop:
The primary change is in `monomer.go`. Everything else comes from that.

NiloCK (approved):
This LGTM.

Small concern: constructing block literals manually is still easy to do, and now results in an invalid state for those blocks. Could possibly add a little godoc color to `monomer.Block` along the lines of:

```
// A Monomer Block. Ethereum compatible header, with cosmos compatible transaction data.
//
// Do not construct manually - instead, use monomer.MakeBlock()

joshklop:
> Could possibly add a little godoc color to monomer.Block along the lines of

Very good idea. Thanks!

PR #70, opened 2024-07-24
Add contract for storing L2 cosmos app state in monomer EVM

natebeauregard:
- Adds a solidity contract and corresponding go bindings to hold the current L2 cosmos app state in contract storage
- This is necessary for implementing withdrawals because we need a valid account proof for the `L2ToL1MessagePasser` via the top-level Monomer state root. The valid account proof is needed for withdrawals and the top-level monomer state root is what the proposer stores in the `L2OutputOracle` contract on L1.
- Adds Makefile targets for easily instaling binding dependencies and for generating bindings
- The `L2ApplicationStateRootProvider` contract will be set to the current Cosmos application `AppHash` every block in a follow-up PR

## Summary by CodeRabbit

- **New Features**
	- Introduced a new contract for managing the application state root in Layer 2 solutions, allowing users to set and update the state root value.
	- Added Go bindings for the Ethereum smart contract, enabling read and write operations, deployment, and session management.
	- Automated script for generating Go bindings for smart contracts, enhancing developer productivity.
	- Enhanced Makefile with new targets for easier installation of dependencies and script execution.

- **Chores**
	- Updated the `.gitignore` to prevent temporary files from being tracked by Git.


contracts/L2ApplicationStateRootProvider.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.25;
```
natebeauregard, line 2: Question to reviewers - what license do we want here? Is `MIT` fine?

bindings/L2ApplicationStateRootProvider.go
natebeauregard, line 1: Question to reviewers - I kept going back and forth on where to put the generated contract bindings and ended up moving them to a new `bindings` package. Let me know if you think they would make more sense in  `/contracts/bindings` or in a different package/directory.

contracts/L2ApplicationStateRootProvider.sol
joshklop: Monomer is Apache 2.0. Let's use that.

bindings/L2ApplicationStateRootProvider.go
joshklop: Location is fine with me. I think we should be able to regenerate the bindings easily though. Wdyt about a Makefile target? Also, we should have a makefile target to install a specific version of abigen. Similar to the other `install-` targets.

joshklop (changes requested):

Makefile
```diff
@@ -45,6 +46,18 @@ install-buf:
 install-go-test-coverage:
 	go install github.com/vladopajic/go-test-coverage/v2@v2.9.0
 
+.PHONY: install-abi-gen
+install-abi-gen:
+	go install github.com/ethereum/go-ethereum/cmd/abigen@v1.10.25
```
natebeauregard, line 51: It would be nice to eventually move all of the specific dependency versions for the `make install...` commands into `versions.json`

contracts/L2ApplicationStateRootProvider.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.25;
+
+/// @title L2ApplicationStateRootProvider
+/// @notice The L2ApplicationStateRootProvider stores the state root of the Cosmos application using Monomer.
+///         This is necessary for implementing withdrawals because we need a valid account proof for the
+///         L2ToL1MessagePasser via the top-level Monomer state root.
+contract L2ApplicationStateRootProvider {
+    /// @notice The current state root of the Cosmos application using Monomer.
+    uint256 public l2ApplicationStateRoot;
```
NiloCK, line 10: In my head, hashed data like state roots are more commonly encoded `bytes32`. Equivalent data, but might save some reformatting work / issues in the future.

scripts/generate-bindings.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+set -e
+
+SCRIPTS_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
```
NiloCK, line 5: `BASH_SOURCE[0]` is new to me. This line is great.

scripts/generate-bindings.sh
```diff
@@ -0,0 +1,59 @@
...
+if ! command -v abigen &> /dev/null
+then
+    echo "abigen could not be found, please run 'make install-abigen'"
+    exit 1
+fi
+if ! command -v forge &> /dev/null
+then
+    echo "forge could not be found, please install 'make install-foundry'"
+    exit 1
+fi
```
NiloCK, line 31: What sort of wonky failure modes might someone encounter if the wrong versions of these tools are present?

Is it worth the effort to be more specific on the checks?

scripts/generate-bindings.sh
```diff
@@ -0,0 +1,59 @@
...
+    # Generate Go bindings using abigen
+    go_binding_file="${BINDINGS_DIR}/${contract_name}.go"
+    abigen --abi=$abi_file --bin $bin_file --pkg="bindings" --type $contract_name --out=$go_binding_file
+
+    echo "Generated Go bindings for ${contract_name}.sol: ${go_binding_file}"
+done
+
+# Remove the temporary forge artifacts and cache directories
+rm -r $FORGE_ARTIFACTS_DIR
+rm -r $FORGE_CACHE_DIR
```
joshklop, line 59: caching is useful sometimes. Let's gitignore these and not remove them

joshklop (approved):

contracts/L2ApplicationStateRootProvider.sol
natebeauregard: Yeah that's a good point, I'll update this!

scripts/install-foundry.sh
```diff
@@ -0,0 +1,67 @@
...
+# Clone the foundry repo temporarily. We do this to avoid the need for a personal access
+# token to interact with the GitHub REST API, and clean it up after we're done.
+git clone https://github.com/foundry-rs/foundry.git "$TMP_DIR" && cd "$TMP_DIR"
+
+# If the nightly tag exists, we can download the pre-built binaries rather than building
+# from source. Otherwise, clone the repository, check out the commit SHA, and build `forge`,
+# `cast`, `anvil`, and `chisel` from source.
+if git rev-parse "$TAG" >/dev/null 2>&1; then
+  echo "Nightly tag exists! Downloading prebuilt binaries..."
+  foundryup -v "$TAG"
```
NiloCK, line 49: Have you run this locally? I have a slight preference toward pinning against a version with an available built binary.

I've had a bad time building forge locally, to the point that I've only reliably succeeded with some custom compiler flags.

contracts/L2ApplicationStateRootProvider.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.25;
+
+/// @title L2ApplicationStateRootProvider
+/// @notice The L2ApplicationStateRootProvider stores the state root of the Cosmos application using Monomer.
+///         This is necessary for implementing withdrawals because we need a valid account proof for the
+///         L2ToL1MessagePasser via the top-level Monomer state root.
+contract L2ApplicationStateRootProvider {
```
NiloCK, line 8: I've only skimmed the op-spec Withdrawals section, but it strikes me as odd that L2 state root storage / update isn't already accounted for somewhere in the OP Stack L1 contracts.

Maybe more clarity on that later today.

NiloCK (approved):
LGTM. Will presumably need a follow-up to include this contract in our e2e L1?

scripts/install-foundry.sh
natebeauregard: Yeah I tested it locally and was able to build `forge` fine. It took around ~20 mins to compile the projects since it uses `cargo --release` mode but it built fine. Would there be any concerns with us pinning forge for monomer to a different version than forge for optimism?

contracts/L2ApplicationStateRootProvider.sol
natebeauregard: The `L2OutputOracle` contract is L1's view of L2 state. This contract will be deployed on an EVM embedded in monomer in a follow-up PR. We need to store this data on monomer so we can eventually generate the account proof (`eth_getProof`) that the withdrawal tx was executed on the cosmos application.

NiloCK: Gotchya.

My understanding was that Cosmos doesn't natively support solidity (seems correct?). This led me to believe this contract was for L1.

Can defer followup until this afternoon.

scripts/generate-bindings.sh
natebeauregard: I investigated this for a little bit and think asserting version would be good to have here, however since there is a good amount of other PRs left to do for the withdrawal implementation I'm going to defer the rest of this investigation until after the rest of the implementation is done.

PR #71, opened 2024-07-25
L1 Deposit Tx

NiloCK:
PR is partial work toward an e2e deposit workflow. It:

- modifies `e2e.Stack.Run` to return some stack data for the test runner
- crafts and submits a deposit tx to the L1 deposit contract, via the op-bindings `portal`
- asserts that the tx landed on L1

The deposit tx is not being processed by Monomer. Likely because monomer's epoch does not advance, in turn likely related to some rollback issues.

To do (future PRs):
- [ ] assert that deposits are picked up from L1 by the node
- [ ] assert minting / balance changes on L2
- [ ] pack a deposit tx with transaction data, processed and executed by monomer

A concern:
- `stack_test.go` is becoming a little unwieldy. May be worth a focused refactor at some point to extract some sort of `action,result` table test so that causes and observable effects can be coupled together more effectively.

## Summary by CodeRabbit

- **New Features**
  - Enhanced the stack's functionality to return detailed configuration information alongside error handling.
  - Introduced a new `StackConfig` structure for better management of configuration parameters.
  - Integrated Layer 1 interactions for deposit transactions in end-to-end tests, including transaction signing and verification.

- **Bug Fixes**
  - Improved error management in the stack setup process.

### DIFF
--- a/e2e/stack.go
+++ b/e2e/stack.go
@@ -34,6 +34,8 @@ type EventListener interface {
 type StackConfig struct {
 	L1URL *url.URL
 	*rollup.Config
+	Operator L1User
+	Users    []L1User
 }
 
 type Stack struct {
@@ -114,6 +116,7 @@ func (s *Stack) Run(ctx context.Context, env *environment.Env) (*StackConfig, er
 	}
 
 	// Fund test EOA accounts
+	l1users := make([]L1User, 0)
 	for range 1 {
 		privateKey, err := crypto.GenerateKey()
 		if err != nil {
@@ -127,7 +130,7 @@ func (s *Stack) Run(ctx context.Context, env *environment.Env) (*StackConfig, er
 			Address: &userAddress,
 		}
 		l1state.OnAccount(user.Address, user)
-		s.L1Users = append(s.L1Users, L1User{
+		l1users = append(l1users, L1User{
 			Address:    userAddress,
 			PrivateKey: privateKey,
 		})
@@ -187,7 +190,7 @@ func (s *Stack) Run(ctx context.Context, env *environment.Env) (*StackConfig, er
 		s.monomerEngineURL,
 		s.opNodeURL,
 		l1Deployments.L2OutputOracleProxy,
-		s.L1Users[0].PrivateKey,
+		l1users[0].PrivateKey,
 		rollupConfig,
 		s.eventListener,
 	)
@@ -196,8 +199,10 @@ func (s *Stack) Run(ctx context.Context, env *environment.Env) (*StackConfig, er
 	}
 
 	return &StackConfig{
-		L1URL:  l1url,
-		Config: rollupConfig,
+		L1URL:    l1url,
+		Config:   rollupConfig,
+		Operator: s.L1Users[0],
+		Users:    l1users[1:],
 	}, nil
 }
 
--- a/e2e/stack.go
+++ b/e2e/stack.go
@@ -47,7 +47,6 @@ type Stack struct {
 	eventListener    EventListener
 	l1BlockTime      uint64
 	prometheusCfg    *config.InstrumentationConfig
-	L1Users          []L1User
 }
 
 type L1User struct {
@@ -117,7 +116,7 @@ func (s *Stack) Run(ctx context.Context, env *environment.Env) (*StackConfig, er
 
 	// Fund test EOA accounts
 	l1users := make([]L1User, 0)
-	for range 1 {
+	for range 2 {
 		privateKey, err := crypto.GenerateKey()
 		if err != nil {
 			return nil, fmt.Errorf("generate key: %v", err)
@@ -201,7 +200,7 @@ func (s *Stack) Run(ctx context.Context, env *environment.Env) (*StackConfig, er
 	return &StackConfig{
 		L1URL:    l1url,
 		Config:   rollupConfig,
-		Operator: s.L1Users[0],
+		Operator: l1users[0],
 		Users:    l1users[1:],
 	}, nil
 }
--- a/e2e/stack.go
+++ b/e2e/stack.go
@@ -13,6 +13,7 @@ import (
 	"github.com/cometbft/cometbft/config"
 	dbm "github.com/cosmos/cosmos-db"
 	opgenesis "github.com/ethereum-optimism/optimism/op-chain-ops/genesis"
+	"github.com/ethereum-optimism/optimism/op-node/rollup"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/crypto"
@@ -28,6 +29,11 @@ import (
 type EventListener interface {
 	OPEventListener
 	node.EventListener
+}
+
+type StackConfig struct {
+	L1URL *url.URL
+	*rollup.Config
 }
 
 type Stack struct {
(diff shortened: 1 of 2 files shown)


e2e/stack_test.go
```diff
@@ -151,6 +213,8 @@ func TestE2E(t *testing.T) {
 		}
 	}
 	t.Log("Monomer blocks contain the l1 attributes deposit tx")
+
+	time.Sleep(15 * time.Second)
```
natebeauregard, line 217: Why is this sleep call necessary now?

e2e/stack_test.go
```diff
@@ -139,6 +196,11 @@ func TestE2E(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, txBlock.Transactions(), 2)
 
+	// inspect L1 for deposit tx receipt
+	receipt, err := l1Client.Client.TransactionReceipt(ctx, depositTx.Hash())
+	require.NoError(t, err, "deposit tx receipt")
+	require.NotNil(t, receipt, "deposit tx receipt")
```
natebeauregard, line 202: Is it worth asserting any information on the deposit tx receipt? It might be nice to verify that the `TransactionDeposited` event was emitted

e2e/stack_test.go
```diff
@@ -99,6 +104,58 @@ func TestE2E(t *testing.T) {
...
+	// send user Deposit Tx
+	nonce, err := l1Client.Client.NonceAt(ctx, user.Address, nil)
+	require.NoError(t, err)
+
+	price, err := l1Client.Client.SuggestGasPrice(context.Background())
+	require.NoError(t, err)
+
+	gasPrice := new(big.Int).Mul(price, big.NewInt(2))
+	if gasPrice.BitLen() > 256 {
+		gasPrice = price // fallback to original price if overflow
```
natebeauregard, line 131: Why do you fallback to the original price here? Should the max gas price just be used instead?

e2e/stack_test.go
```diff
@@ -99,6 +104,58 @@ func TestE2E(t *testing.T) {
...
+			},
+			Nonce:    big.NewInt(int64(nonce)),
+			GasPrice: big.NewInt(price.Int64() * 2),
+			GasLimit: 1e6,
+			Value:    big.NewInt(1e18 / 10), // 0.1 eth
+			Context:  ctx,
+			NoSend:   false,
+		},
+		user.Address,
+		big.NewInt(1e18/20), // 0.5 eth deposit to L2
```
natebeauregard, line 152: Readbility might be better if you extract `big.NewInt(1e18/20)` out to a contant

e2e/stack_test.go
```diff
@@ -99,6 +104,58 @@ func TestE2E(t *testing.T) {
...
+			From: user.Address,
+			Signer: func(addr common.Address, tx *types.Transaction) (*types.Transaction, error) {
+				signed, err := types.SignTx(tx, signer, user.PrivateKey)
+				if err != nil {
+					return nil, err
+				}
+				return signed, nil
+			},
+			Nonce:    big.NewInt(int64(nonce)),
+			GasPrice: big.NewInt(price.Int64() * 2),
```
natebeauregard, line 145: Could you use the `gasPrice` variable you initialized above here?

e2e/stack_test.go
```diff
@@ -99,6 +104,58 @@ func TestE2E(t *testing.T) {
...
+			GasLimit: 1e6,
```
natebeauregard, line 146: Why is `1e6` used here for gas limit? I see that the gas limit is also passed in as `stackConfig.Genesis.SystemConfig.GasLimit/10` down below

natebeauregard (commented):
Changes look real good! Left a couple questions for my own understanding but I'll approve once they're answered

e2e/stack.go
```diff
@@ -121,74 +129,80 @@ func (s *Stack) Run(ctx context.Context, env *environment.Env) error {
...
+		return nil, fmt.Errorf("run the op stack: %v", err)
 	}
-	return nil
+
+	return &StackConfig{
+		L1URL:    l1url,
+		Config:   rollupConfig,
+		Operator: l1users[0],
+		Users:    l1users[1:],
+	}, nil
```
joshklop, line 205: I haven't looked through the PR in detail yet, but my impression is that this can be quite dangerous. The caller will be able to modify the data provided in this `StackConfig`. Probably better for the caller to pass in this kind of information to the `Stack`.

joshklop: On second thought, my points don't make much sense. I'll let Nate take the reviews here.

e2e/stack_test.go
NiloCK: This suggestion bore fruit. As it stood, the transaction was reverting. Event verification added here: 5253de9. Adjustment to MaxGas that fixed the revert in the prior commit.

e2e/stack_test.go
NiloCK: b169ea0, ec36f24

These are mostly dummy values. I extracted `oneEth` and used it directly.

e2e/stack_test.go
NiloCK: Covered in https://github.com/polymerdao/monomer/commit/c65a3dfb7478eb3edec8e6b74b2d09db9dc90638. Redundant gas price calculation removed.

e2e/stack_test.go
NiloCK: Mostly trial and error. The `1e6` here is the gasLimit for the L1 tx. It has since changed to `1e7`, which fixes a revert.

The other gasLimit is for L2 processing of the deposit transaction.

natebeauregard: Ah ok thanks. Could you pull it out into a constant for readability?

e2e/stack_test.go
```diff
@@ -149,8 +224,12 @@ func TestE2E(t *testing.T) {
 			require.NoError(t, err)
 			require.Fail(t, fmt.Sprintf("expected tx to be deposit tx: %s", txBytes))
 		}
+		if len(txs) > 1 {
+			t.Logf("%d txs", len(txs))
+		}
 	}
 	t.Log("Monomer blocks contain the l1 attributes deposit tx")
+	// time.Sleep(15 * time.Second)
```
natebeauregard, line 232: Can you remove this comment before merge?

natebeauregard (approved):
LGTM! Left a couple small suggestions for a constant to be extracted and for a comment to be removed but no need to re-review before squash/merge

e2e/stack_test.go
NiloCK: Refactored in 71d7221, but there's the interplay between L1 and L2 gasLimit values that discourages me from pulling either as a constant.

(L1 gas gets burned according to the l2Gaslimit as part of the cross-chain gas accounting.)

e2e/stack_test.go
NiloCK: Always sneaks back in. I just love sleeping for 15 seconds.

e2e/stack.go
NiloCK: Third thought - I think these points do make sense. The returned data should be "information only", and not knobs of the running stack that can be twiddled by the caller.

Tracked in [notion](https://www.notion.so/nethermind/e2e-return-copied-data-from-stack-Run-db0e763ab0d8489598574f718e3d8edd?pvs=4).

PR #72, opened 2024-07-25
Add ethereum state trie to monomer

natebeauregard:
- Passes an ethereum state DB into monomer and creates a state trie at the root hash
- Updates the monomer block hash to be the ethereum state root
- In follow-up PRs this ethereum state trie will be updated whenever withdrawal transactions are initiated and will be verified by `op-node` when proving that the withdrawal transaction took place on monomer

## Summary by CodeRabbit

- **New Features**
  - Enhanced integration of Ethereum state management, allowing improved cross-chain capabilities.
  - New in-memory Ethereum state database created for testing and operational purposes.

- **Bug Fixes**
  - Updated method signatures and internal logic to accommodate new Ethereum state parameters, ensuring compatibility with existing functionality.

- **Tests**
  - Expanded test framework to simulate Ethereum state management scenarios, improving the accuracy of transactions and state handling.


builder/builder.go
```diff
@@ -127,12 +132,15 @@ func (b *Builder) Build(ctx context.Context, payload *Payload) (*monomer.Block,
...
+	cosmosAppHash := info.GetLastBlockAppHash()
+	ethStateRoot := b.ethStateTrie.Hash().Bytes()
+	monomerAppHash := crypto.Keccak256(cosmosAppHash, ethStateRoot)
 	header := &monomer.Header{
 		ChainID:    b.chainID,
 		Height:     currentHeight + 1,
 		Time:       payload.Timestamp,
 		ParentHash: currentHead.Header.Hash,
-		AppHash:    info.GetLastBlockAppHash(),
+		AppHash:    monomerAppHash,
```
joshklop, line 143: Let's avoid variables only used one time.
```go
AppHash: crypto.Keccak256(info.GetLastBlockAppHash(), b.ethStateTrie.Hash().Bytes())
```

testutils/utils.go
```diff
@@ -36,11 +38,29 @@ func NewMemDB(t *testing.T) dbm.DB {
...
+		require.NoError(t, rawstatedb.Close())
+	})
+	ethstatedb, err := state.New(gethtypes.EmptyRootHash, state.NewDatabase(rawstatedb), nil)
+	require.NoError(t, err)
+	return ethstatedb
+}
+
+func NewEthStateTrie(t *testing.T) *trie.StateTrie {
+	statedb := NewEthStateDB(t)
+	ethStateTrieID := trie.StateTrieID(gethtypes.EmptyRootHash)
```
joshklop, line 53: same as above, we can avoid the one-time use variable

e2e/stack.go
```diff
@@ -211,6 +213,12 @@ func (s *Stack) runMonomer(ctx context.Context, env *environment.Env, genesisTim
 	env.DeferErr("close tx db", txdb.Close)
 	mempooldb := dbm.NewMemDB()
 	env.DeferErr("close mempool db", mempooldb.Close)
+	rawstatedb := rawdb.NewMemoryDatabase()
```
joshklop, line 216: one-time var

genesis/genesis.go
```diff
@@ -21,7 +21,7 @@ type Genesis struct {
 const defaultGasLimit = 30_000_000
 
 // Commit assumes the application has not been initialized and that the block store is empty.
-func (g *Genesis) Commit(ctx context.Context, app monomer.Application, blockStore store.BlockStoreWriter) error {
+func (g *Genesis) Commit(ctx context.Context, app monomer.Application, blockStore store.BlockStoreWriter, ethStateRoot []byte) error {
```
joshklop, line 24: prefer descriptive types
```suggestion
func (g *Genesis) Commit(ctx context.Context, app monomer.Application, blockStore store.BlockStoreWriter, ethStateRoot common.Hash) error {
```

node/node.go
```diff
@@ -56,6 +60,7 @@ func New(
 	blockdb,
 	mempooldb dbm.DB,
 	txdb cometdb.DB,
+	ethstatedb *state.StateDB,
```
joshklop, line 63: Can this be an interface instead? (apologies if we discussed this on another thread and I forgot)

builder/builder.go
```diff
@@ -10,19 +10,22 @@ import (
...
-	txStore    txstore.TxStore
-	eventBus   *bfttypes.EventBus
-	chainID    monomer.ChainID
+	mempool      *mempool.Pool
+	app          monomer.Application
+	blockStore   store.BlockStore
+	txStore      txstore.TxStore
+	eventBus     *bfttypes.EventBus
+	chainID      monomer.ChainID
+	ethStateTrie *trie.StateTrie
```
joshklop, line 28: It's a bit confusing (I'd like to change it in the future), but the builder controls rollbacks. So we need to rollback the state trie in `Rollback`

joshklop (changes requested):

e2e/stack.go
natebeauregard: we need this as a separate var so we can [close it](https://github.com/polymerdao/monomer/pull/72/files/e93ca84aec578d548311912b0870aa318b2074c6#diff-a60e40fd7c6abc129086b0e1217ee5295a4e4cbff6510b6f5fd253faaad6b553R221) after the test runs

joshklop: Got it, thanks. Can we move the defer right after this assignment? That way if we hit the error case, we still close the db.

node/node.go
joshklop: Like this https://github.com/ethereum/go-ethereum/blob/ac0f220040bd4271a040170bc0b039eb88dee9bd/core/vm/interface.go#L32

joshklop: Nvm, just saw your comment in the other github thread about obtaining the trie and responded to it. I like this `*state.StateDB` now :)

NiloCK:
Init question @natebeauregard: is the added Etherume State Trie _only_ responsible for withdrawals?

builder/builder.go
NiloCK: Similar: `get_Proof` api wants to be callable by block hash / number. As far as I can tell, this implementation doesn't store historical state trie data, so those get proof calls can't be computed other than at the current block with the in-memory trie.

natebeauregard: @NiloCK Yeah that's a good point. I'm currently working on making the DB snapshots indexable by block root to support rollbacks so as long as we have the block hash/number we should be able to get the block root off of the header to use for `get_proof`.

monomer.go
```diff
@@ -50,18 +49,17 @@ type Header struct {
 	Height     int64       `json:"height"`
 	Time       uint64      `json:"time"`
 	ParentHash common.Hash `json:"parentHash"`
-	// state after txs from the *previous* block
-	AppHash  []byte      `json:"app_hash"`
-	GasLimit uint64      `json:"gasLimit"`
-	Hash     common.Hash `json:"hash"`
+	StateRoot  common.Hash `json:"stateRoot"`
```
natebeauregard, line 52: Note to reviewers - the `AppHash` field on the monomer header was repurposed to directly store the ethereum state root

natebeauregard:
Had to force push after a messy rebase :cry:

builder/builder.go
```diff
@@ -153,6 +157,17 @@ func (b *Builder) Build(ctx context.Context, payload *Payload) (*monomer.Block,
 		return nil, fmt.Errorf("commit: %v", err)
 	}
 
+	ethState, err := state.New(b.blockStore.HeadBlock().Header.StateRoot, b.ethstatedb, nil)
+	if err != nil {
+		return nil, fmt.Errorf("create ethereum state: %v", err)
+	}
+	// TODO: execute withdrawals
+	ethStateRoot, err := ethState.Commit(uint64(currentHeight+1), true)
```
joshklop, line 165: to avoid counting twice
```suggestion
	ethStateRoot, err := ethState.Commit(uint64(header.Height), true)
```

builder/builder_test.go
```diff
@@ -208,7 +211,7 @@ func TestRollback(t *testing.T) {
 		require.NoError(t, eventBus.Stop())
 	})
 
-	require.NoError(t, g.Commit(context.Background(), app, blockStore))
+	require.NoError(t, g.Commit(context.Background(), app, blockStore, ethstatedb))
```
joshklop, line 214: Let's add todos to add more tests to ensure we're actually updating the withdrawals trie.

genesis/genesis_test.go
```diff
@@ -48,8 +50,9 @@ func TestCommit(t *testing.T) {
 				require.NoError(t, blockstoredb.Close())
 			})
 			blockStore := store.NewBlockStore(blockstoredb)
+			ethstatedb := testutils.NewEthStateDB(t)
```
joshklop, line 53: Sam here, adding todos for more tests. It's fine for now.

joshklop (approved):

joshklop (approved):

PR #73, opened 2024-07-25
Split NewBlock and SetHeader out from MakeBlock

joshklop:
This will be useful in cases where we don't need to recalculate the hash.


natebeauregard (approved):
LGTM

PR #75, opened 2024-07-26
Decrease Sequencer Window Size

joshklop:
Right now, the default SWS is 300, which is way too large for our e2e tests that only generate a few L1 blocks. By setting the SWS to the minimum value (two), the op-node will now perform [unsafe block consolidation].

[unsafe block consolidation]: https://specs.optimism.io/glossary.html#unsafe-block-consolidation


natebeauregard (approved):
LGTM

joshklop:
Argh, rollback issues affecting the e2e test. Will examine.

node/node.go
```diff
@@ -110,7 +111,7 @@ func (n *Node) Run(ctx context.Context, env *environment.Env) error {
 				*eth.Block
 			}{
 				ChainID: eth.NewChainID(n.genesis.ChainID.HexBig(), ethMetrics),
-				Block:   eth.NewBlock(blockStore, ethMetrics),
+				Block:   eth.NewBlock(blockStore, new(big.Int).SetUint64(uint64(n.genesis.ChainID)), ethMetrics),
```
natebeauregard, line 114: It might be nice to add a `Big()` func to `ChainID` that would just return `new(big.Int).SetUint64(uint64(id))`. Then, the `ChainID.HexBig` func could be refactored to use `ChainID.Big` internally too

e2e/stack.go
```diff
@@ -91,7 +90,14 @@ func (s *Stack) Run(ctx context.Context, env *environment.Env) (*StackConfig, er
 		return nil, fmt.Errorf("new deploy config: %v", err)
 	}
 	deployConfig.SetDeployments(l1Deployments)
+	deployConfig.L1GenesisBlockTimestamp = hexutil.Uint64(time.Now().Unix())
 	deployConfig.L1UseClique = false // Allows node to produce blocks without addition config. Clique is a PoA config.
+	deployConfig.SequencerWindowSize = 4
```
natebeauregard, line 95: could you add a brief comment here for transparency describing why the sequencer window size needs to be set as 4?

natebeauregard (approved):
LGTM. Left a couple nits but no need to re-review

e2e/gethdevnet.go
```diff
@@ -33,6 +33,7 @@ func gethdevnet(blockTime uint64, genesis *core.Genesis) (*rpc.Client, string, e
 	if err != nil {
 		return nil, "", fmt.Errorf("start geth L1: %w", err)
 	}
+	// TODO close node
```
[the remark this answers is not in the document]
joshklop: Yep, good reminder. Thanks!

PR #76, opened 2024-07-29
Adapters-unit-tests

AnkushinDaniil:
Please check the current TODOs
There are some errors I can't emulate and I'm not sure this is necessary.
See TODO about AdaptCosmosTxsToEthTxs. There is an error when trying to compare the whole transaction (rollupCostData.v is different). There is a possibility that AdaptCosmosTxsToEthTxs does not work correctly. Please give me feedback.

## Summary by CodeRabbit

- **New Features**
	- Introduced a comprehensive suite of unit tests to validate the conversion of Ethereum transactions to Cosmos transactions and vice versa.
	- Enhanced clarity in code with explicit package aliasing for improved readability.
	- Added new functions for converting Ethereum transactions into Cosmos transactions, improving modularity.

- **Bug Fixes**
	- Improved robustness by ensuring the adaptation functions handle erroneous inputs and edge cases effectively.

### DIFF
--- a/x/rollup/types/adapters_test.go
+++ b/x/rollup/types/adapters_test.go
@@ -105,7 +105,7 @@ func TestAdaptPayloadTxsToCosmosTxs(t *testing.T) {
 		t.Run("unmarshal binary error", func(t *testing.T) {
 			cosmosTxs, err := rolluptypes.AdaptPayloadTxsToCosmosTxs([]hexutil.Bytes{[]byte("invalid")})
 			require.Nil(t, cosmosTxs)
-			require.ErrorContains(t, err, "unmarshal binary")
+			require.Error(t, err)
 		})
 		t.Run("zero deposit txs", func(t *testing.T) {
 			inner := generateDynamicFeeInner(r)
@@ -137,7 +137,7 @@ func TestAdaptPayloadTxsToCosmosTxs(t *testing.T) {
 
 			cosmosTxs, err := rolluptypes.AdaptPayloadTxsToCosmosTxs([]hexutil.Bytes{depTxBytes, nonDepTxBytes, []byte("invalid")})
 			require.Nil(t, cosmosTxs)
-			require.ErrorContains(t, err, "unmarshal binary tx: ")
+			require.Error(t, err)
 		})
 	})
 }
--- a/x/rollup/types/adapters_test.go
+++ b/x/rollup/types/adapters_test.go
@@ -1,7 +1,6 @@
 package types_test
 
 import (
-	"fmt"
 	"math/big"
 	"testing"
 
@@ -13,6 +12,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	rollupv1 "github.com/polymerdao/monomer/gen/rollup/v1"
 	rolluptypes "github.com/polymerdao/monomer/x/rollup/types"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
@@ -24,6 +24,7 @@ import (
 // deposit transaction. The resulting deposit transaction is then compared to the
 // original one to ensure that the conversion process was successful.
 func TestAdaptPayloadTxsToCosmosTxs(t *testing.T) {
+	// TODO: create an inner generator function to generate the test data
 	// Define the necessary transaction parameters.
 	sourceHashString := "0x1234567890abcdef1234567890abcdef12345678"
 	fromAddressString := "0x1111111111111111111111111111111111111111"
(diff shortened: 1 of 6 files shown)


AnkushinDaniil:
@joshklop 
review request

joshklop (changes requested):
Good start. Can we use these helpers to generate the eth txs? https://github.com/polymerdao/monomer/blob/8cba2d2b12672a2937eb34d3d59e604051c2aa90/testutils/utils.go#L41

AnkushinDaniil:
Yes, we can, but I decided to write less hard-coded function to increase the variety of transactions by using more random data. I'll change it if necessary. Or we can use my variants instead of this function. Which do you prefer?

AnkushinDaniil:
Proof analysis:
Current adapters don't need optimisation

x/rollup/types/adapters_test.go
```diff
@@ -0,0 +1,308 @@
...
+	rolluptypes "github.com/polymerdao/monomer/x/rollup/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAdaptPayloadTxsToCosmosTxs(t *testing.T) {
+	t.Run("Zero txs", func(t *testing.T) {
+		cosmosTxs, err := rolluptypes.AdaptPayloadTxsToCosmosTxs([]hexutil.Bytes{})
+		require.NoError(t, err)
+		assert.Equal(t, 0, len(cosmosTxs))
```
joshklop, line 26: `require`. I think there is `require.Empty`

x/rollup/types/adapters_test.go
```diff
@@ -0,0 +1,308 @@
...
+			nonDepTx := ethtypes.NewTx(nonDepInner)
+			nonDepTxBytes, err := nonDepTx.MarshalBinary()
+			require.NoError(t, err)
+
+			cosmosTxs, err := rolluptypes.AdaptPayloadTxsToCosmosTxs([]hexutil.Bytes{depTxBytes, nonDepTxBytes, []byte("invalid")})
+			assert.Nil(t, cosmosTxs)
+			assert.ErrorContains(t, err, "unmarshal binary tx: ")
+		})
+	})
+}
```
joshklop, line 148: Let's not test the precise error conditions. That makes the tests too dependent on the implementation. The tests should be run from the perspective of someone who only expects a certain behavior and does not know about the implementation. We should be able to refactor the implementation without having to change the tests.

x/rollup/types/adapters_test.go
```diff
@@ -0,0 +1,308 @@
...
+			{
+				name:      "DepositTx + AccessListTx",
+				depNum:    1,
+				nonDepNum: 1,
+			},
+			{
+				name:      "Multiple DepositTxs + DynamicFeeTxs",
+				depNum:    3,
+				nonDepNum: 3,
+			},
```
joshklop, line 54: I really like how you test various real-world scenarios here.

x/rollup/types/adapters_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+				depositTxsNum := 0
+				for i := range transactions {
+					txBinary, err := transactions[i].MarshalBinary()
+					require.NoError(t, err)
+					ethTxs[i] = txBinary
+					if transactions[i].IsDepositTx() {
+						depositTxsNum++
+					}
+				}
```
joshklop, line 74: This looks a lot like the real implementation... are we overfitting for the implementation here?

joshklop (changes requested):

x/rollup/types/adapters_test.go
```diff
@@ -0,0 +1,303 @@
...
+			transaction := ethtypes.NewTx(inner)
+			txBytes, err := transaction.MarshalBinary()
+			require.NoError(t, err)
+			cosmosTxs, err := rolluptypes.AdaptPayloadTxsToCosmosTxs([]hexutil.Bytes{txBytes})
+			require.Nil(t, cosmosTxs)
+			require.Error(t, err)
+		})
+		t.Run("NewAnyWithValue error", func(t *testing.T) {
+			t.Skip()
+			// TODO: Implement this test case
```
joshklop, line 121: Is this PR ready for review?

joshklop (changes requested):

x/rollup/types/adapters_test.go
AnkushinDaniil: Yes, I am just not really sure if we really need to implement these test cases.

x/rollup/types/adapters_test.go
```diff
@@ -0,0 +1,303 @@
...
+					ethTxs[0].SetTime(adoptedTxs[0].Time())
+					require.Equal(t, ethTxs[i].Data(), adoptedTxs[i].Data())
+					// TODO: Incorrect adaptation of other fields
+				}
+			})
+		}
+	})
+}
+
+func generateCosmosSDKTx(depTxsNum, nonDepTxsNum int, ethTxs []*ethtypes.Transaction) bfttypes.Txs {
```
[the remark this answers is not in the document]
joshklop: > it fails in case of using testapp.ToTx

Why?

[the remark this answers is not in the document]
AnkushinDaniil: The data being unmarshaled [here](https://github.com/AnkushinDaniil/monomer/blob/69a101210e3b37119de05de17d0488b357197c97/x/rollup/types/adapters.go#L113) should always have been created by the `MarshalBibary()` function.
When we adapt transactions created by `testapp.ToTx` we get an error: `"transaction type not supported"` in [here](https://github.com/AnkushinDaniil/monomer/blob/69a101210e3b37119de05de17d0488b357197c97/x/rollup/types/adapters.go#L113)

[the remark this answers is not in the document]
AnkushinDaniil: I don't know if it is possible to use `testapp.ToTx` here. I have to rewrite all tests to use testapp and testutils and realism and variety of tests will decrease in this case. What do you think @joshklop?

PR #81, opened 2024-07-30
`eth.get_proof` - Consume data sources from node

NiloCK:
PR updates `ProofProvider.getState()` from a noop to a tentatively working implementation.

Tentative because the existing test remains a stub - a failure assertion on a call to a partially constructed `ProofProvider`.

I'm inclined to review and merge this ASAP, so that unit and e2e tests can be started in parallel.

## Summary by CodeRabbit

- **New Features**
  - Enhanced the ProofProvider to interact more effectively with state databases and block stores, improving its capabilities for blockchain state management.
  - Implemented error handling in the ToEth method, ensuring graceful handling of nil block instances.

- **Bug Fixes**
  - Updated the getState method to retrieve the latest block data, transitioning from a placeholder to a fully functional implementation.

- **Refactor**
  - Improved the initialization of the ProofProvider and its parameters, refining how it accesses state data.
  
- **Style**
  - Made minor corrections to variable naming for better clarity and accuracy.


eth/get_proof.go
```diff
@@ -33,17 +36,41 @@ func (n *proofList) Delete(key []byte) error {
...
+		database:   db,
+		blockStore: blockStore,
+	}
 }
 
 // getState returns the state.StateBD and Header of block at the given number.
 // If the passed number is nil, it returns the the db and header of the latest block.
 //
-// TODO: replace this with a real implementation based on `ProofProvider` being supplied
+// WIP: replace this with a real implementation based on `ProofProvider` being supplied
```
natebeauregard, line 49: should this `WIP` comment be removed? Or is to show that the impl may change once the unit test is updated to be more than a stub?

eth/get_proof.go
```diff
@@ -33,17 +36,41 @@ func (n *proofList) Delete(key []byte) error {
...
+		return nil, types.Header{}, fmt.Errorf("getting eth block %d: %w", blockNumber, err)
+	}
+
+	header := ethBlock.Header()
+	hash := ethBlock.Hash()
+
+	sdb, err := state.New(hash, p.database, nil)
+	if err != nil {
+		return nil, *header, fmt.Errorf("opening state.StateDB: %w", err)
+	}
```
natebeauregard, line 71: Since the ethereum `StateRoot` is stored on the monomer block header, the block shouldn't need to be converted to an eth block. I think we can just pull `StateRoot` directly off of the monomer block header and use it to build the eth state 

```suggestion
	var block *types.Block
	var err error

	if blockNumber == nil {
		block, err = p.blockStore.HeadBlock()
	} else {
		block, err = p.blockStore.BlockByNumber(blockNumber.Int64())
	}

	if err != nil {
		return nil, types.Header{}, fmt.Errorf("getting block %d: %w", blockNumber, err)
	}

	header := block.Header()

	sdb, err := state.New(header.StateRoot, p.database, nil)
	if err != nil {
		return nil, *header, fmt.Errorf("opening state.StateDB: %w", err)
	}
```

eth/get_proof.go
```diff
@@ -181,7 +208,7 @@ func (p *ProofProvider) getProof(address common.Address, storageKeys []string, b
 		var storageTrie state.Trie
 		if storageRoot != types.EmptyRootHash && storageRoot != (common.Hash{}) {
 			id := trie.StorageTrieID(header.Root, crypto.Keccak256Hash(address.Bytes()), storageRoot)
-			st, err := trie.NewStateTrie(id, statedb.Database().TrieDB())
+			st, err := trie.NewStateTrie(id, statedb.Database().TrieDB()) // todo
```
natebeauregard, line 211: curious - what's this TODO comment for?

eth/eth_test.go
```diff
@@ -128,10 +128,11 @@ func TestGetBlockByHash(t *testing.T) {
...
+	blockstore := store.NewBlockStore(testutils.NewMemDB(t)) // this blockstore contains no blocks.
 
-	proofProvider := eth.NewProofProvider()
+	proofProvider := eth.NewProofProvider(nil, blockstore)
 
 	pf, err := proofProvider.GetProof(someAddress, []string{}, nil)
-	require.ErrorIs(t, err, eth.ErrNotImplemented, "getProof didn't fail")
-	require.Nil(t, pf, "received proof")
+	require.Error(t, err, "should not succeed in generating proofs with empty blockstore")
+	require.Nil(t, pf, "received proof from empty blockstore")
```
natebeauregard, line 137: This is fine for now, I agree that we'll want to add more comprehensive unit tests for this later on though. Should I create a subtask in the withdrawals page for adding this test or did you already have something planned?

eth/get_proof.go
NiloCK: Yup. 9c0fe5a

eth/get_proof.go
NiloCK: The `header` produced in this change is a `monomer.Header` but the (copied) code in `GetProof` is wanting a `geth.core.types.Header`.

Consuming code only really wants the `Root` from this header, so I could change to return that directly from the monomer header, but I'm more inclined to minimize the diff in the copied functions, at least until some tests are in place.

eth/eth_test.go
NiloCK: Could create a subtask. I mostly wanted to checkpoint before investing much time in creating the "wrong" types of testdata.

eth/get_proof.go
natebeauregard: Ah ok I see your point, we can always revisit this during testing. In that case could you use `ethBlock.Root()` instead of `ethBlock.Hash()`? I think that has the eth state root that we want to use in the state db.

joshklop: > I'm more inclined to minimize the diff in the copied functions, at least until some tests are in place.

Agreed here. Helps with future-proofing too :+1:

natebeauregard: Sounds good, I'll go ahead and approve for now and then we can revisit this tomorrow once we have the withdrawals in eth storage

natebeauregard (approved):

PR #82, opened 2024-07-30
Add signer field to ApplyL1Txs protobuf definition

NiloCK:


NiloCK:
Similar changes were merged with #123.

PR #83, opened 2024-07-30
Add requirements installs to `make install-buf`

NiloCK:
The `install-buf` command didn't previously add all the requirements to successfully run `make gen-proto`.

PR adds installation of two protobuf plugins.

## Summary by CodeRabbit

- **New Features**
	- Enhanced the installation process for development tools related to the Cosmos ecosystem, improving support for protocol buffer generation.
	- Integrated new commands to install necessary plugins for a more efficient development experience.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,8 @@ install-gofumpt:
 .PHONY: install-buf
 install-buf:
 	go install github.com/bufbuild/buf/cmd/buf@v1.31.0
+	go install github.com/cosmos/gogoproto/protoc-gen-gocosmos@1.5.0
+	go install github.com/cosmos/cosmos-proto/cmd/protoc-gen-go-pulsar@1.0.0-beta.5
 
 .PHONY: install-go-test-coverage
 install-go-test-coverage:


joshklop:
Does `buf mod tidy` take care of this? I think `buf` manages plugins itself, but I can't recall exactly.

NiloCK:
> Does `buf mod tidy` take care of this? I think `buf` manages plugins itself, but I can't recall exactly.

`buf mod tidy` doesn't seem to be a command. `buf mod update` might, if we had `buf.yaml / buf.lock` files configured, but our generate command uses the `buf.gen.gocosmos.yml` file instead of a "locally configured buf project" (guessing at vocabulary here).

The error message when a plugin isn't found doesn't suggest any `buf` specific remedies. It just complains about a missing binary:

```
~/n/p/monomer (cmk.applyL1Tx-signer)$ make gen-proto 
$(go env GOPATH)/bin/buf generate --template buf.gen.gocosmos.yaml --exclude-path proto/rollup/module,testapp/proto/testapp/module
Failure: plugin gocosmos: could not find protoc plugin for name gocosmos - please make sure protoc-gen-gocosmos is installed and present on your $PATH
```

joshklop (approved):
I'm convinced. Thanks!

NiloCK:
> I'm convinced. Thanks!

I think there's probably a `protobuffier` way to do this, but the script here is simple and encapsulated enough to not cause too many problems.

PR #85, opened 2024-07-30
Use a persistent pebble db for the eth state

natebeauregard:
Use a persistent pebble db for the eth state

## Summary by CodeRabbit

- **New Features**
	- Improved database management by transitioning from an in-memory database to a more reliable persistent storage solution for Ethereum state data.
	- Enhanced error handling during database initialization, providing clearer feedback in case of issues.

- **Bug Fixes**
	- Fixed potential issues related to data persistence and retrieval with the new database implementation.

### DIFF
--- a/integrations/integrations.go
+++ b/integrations/integrations.go
@@ -29,6 +29,8 @@ import (
 
 const (
 	monomerEngineWSFlag = "monomer-engine-ws"
+	defaultCacheSize    = 16 // 16 MB
+	defaultHandlesSize  = 16
 )
 
 var sigCh = make(chan os.Signal, 1)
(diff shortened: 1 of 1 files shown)


integrations/integrations.go
```diff
@@ -29,6 +29,8 @@ import (
 
 const (
 	monomerEngineWSFlag = "monomer-engine-ws"
+	defaultCacheSize    = 16 // 16 MB
+	defaultHandlesSize  = 16
```
natebeauregard, line 33: Any opinions on what these defaults should be?

NiloCK: Seems reasonable enough.

Do we have a strategy for handling configuration?

NiloCK (approved):
:+1:

integrations/integrations.go
natebeauregard: I reused the `RootDir` specified by the user in `svrCtx.Config`, otherwise I'd say the rest of the config is probably just going to be abstracted away to monomer. We can always revisit this if users want more control over their eth state db instance

PR #87, opened 2024-07-31
Add x/rollup withdrawal msg server

natebeauregard:
- Adds an `InitiateWithdrawal` msg server to the `x/rollup` module
- Burns user ETH tokens through the mint module when a withdrawal is initiated
- Some extra information is passed through the `InitiateWithdrawalRequest` message that's not currently used in the rollup module (`target`, `gas_limit`, `data`), however it will be used for processing in the monomer EVM in a subsequent PR
- Does some housekeeping in the rollup module msg server for displaying the correct error messages and passing through the correct context

## Summary by CodeRabbit

- **New Features**
  - Introduced a new method for burning ETH from user accounts.
  - Added functionality for enhanced logging during the withdrawal process.
  - Expanded event handling with new types for burning ETH and initiating withdrawals.
  
- **Bug Fixes**
  - Improved error handling for invalid sender addresses during ETH transactions.
  
- **Chores**
  - Added new error codes to enhance error management related to Ethereum operations.

### DIFF
--- a/x/rollup/keeper/msg_server.go
+++ b/x/rollup/keeper/msg_server.go
@@ -35,47 +35,47 @@ func (k *Keeper) ApplyL1Txs(goCtx context.Context, msg *rollupv1.ApplyL1TxsReque
 	}
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
-	ctx.Logger().Debug("processing L1 txs", "txCount", len(msg.TxBytes))
+	ctx.Logger().Debug("Processing L1 txs", "txCount", len(msg.TxBytes))
 
 	// process L1 system deposit tx
 	txBytes := msg.TxBytes[0]
 	var tx ethtypes.Transaction
 	if err := tx.UnmarshalBinary(txBytes); err != nil {
-		ctx.Logger().Error("failed to unmarshal system deposit transaction", "index", 0, "err", err, "txBytes", txBytes)
+		ctx.Logger().Error("Failed to unmarshal system deposit transaction", "index", 0, "err", err, "txBytes", txBytes)
 		return nil, types.WrapError(types.ErrInvalidL1Txs, "failed to unmarshal system deposit transaction: %v", err)
 	}
 	if !tx.IsDepositTx() {
-		ctx.Logger().Error("first L1 tx must be a system deposit tx", "type", tx.Type())
+		ctx.Logger().Error("First L1 tx must be a system deposit tx", "type", tx.Type())
 		return nil, types.WrapError(types.ErrInvalidL1Txs, "first L1 tx must be a system deposit tx, but got type %d", tx.Type())
 	}
 	l1blockInfo, err := derive.L1BlockInfoFromBytes(k.rollupCfg, 0, tx.Data())
 	if err != nil {
-		ctx.Logger().Error("failed to derive L1 block info from L1 Info Deposit tx", "err", err, "txBytes", txBytes)
+		ctx.Logger().Error("Failed to derive L1 block info from L1 Info Deposit tx", "err", err, "txBytes", txBytes)
 		return nil, types.WrapError(types.ErrInvalidL1Txs, "failed to derive L1 block info from L1 Info Deposit tx: %v", err)
 	}
 
 	// save L1 block info to AppState
 	if err := k.SetL1BlockInfo(ctx, *l1blockInfo); err != nil {
-		ctx.Logger().Error("failed to save L1 block info to AppState", "err", err)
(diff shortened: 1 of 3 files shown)


x/rollup/keeper/msg_server.go
```diff
@@ -130,7 +157,37 @@ func (k *Keeper) MintETH(ctx *sdk.Context, addr sdk.AccAddress, amount sdkmath.I
...
+	coins := sdk.NewCoins(sdk.NewCoin(types.ETH, amount))
+
+	// Transfer the coins to withdraw from the user account to the mint module
+	err := k.bankkeeper.SendCoinsFromAccountToModule(ctx, addr, types.MintModule, coins)
+	if err != nil {
+		return fmt.Errorf("failed to send withdrawal coins from user account %v to mint module: %v", addr, err)
+	}
+
+	// Burn the ETH coins from the mint module
+	if err := k.bankkeeper.BurnCoins(ctx, types.MintModule, coins); err != nil {
```
natebeauregard, line 177: I chose to burn the coins from the mint module, however they could also be burned (or even minted for deposits) through the bank module too. Lmk if you have any preference which module we go through

joshklop: Bank keeper is probably better. Mint module [appears to be](https://docs.cosmos.network/main/build/modules/mint) mainly for calculating staking rewards and token inflation, which is what Monomer helps avoid.

x/rollup/keeper/msg_server.go
```diff
@@ -140,7 +197,7 @@ func (k *Keeper) MintETH(ctx *sdk.Context, addr sdk.AccAddress, amount sdkmath.I
 //
 // Persisted data conforms to optimism specs on L1 attributes:
 // https://github.com/ethereum-optimism/optimism/blob/develop/specs/deposits.md#l1-attributes-predeployed-contract
-func (k *Keeper) SetL1BlockInfo(ctx *sdk.Context, info derive.L1BlockInfo) error { //nolint:gocritic
+func (k *Keeper) SetL1BlockInfo(ctx sdk.Context, info derive.L1BlockInfo) error { //nolint:gocritic
```
joshklop, line 200: Purpose for this change? Not saying it isn't good, just curious

x/rollup/keeper/msg_server.go
```diff
@@ -95,31 +96,57 @@ func (k *Keeper) ApplyL1Txs(goCtx context.Context, msg *rollupv1.ApplyL1TxsReque
...
+	ctx.Logger().Debug("withdrawing L2 assets", "sender", msg.Sender, "amount", msg.Amount)
+
+	cosmAddr, err := sdk.AccAddressFromBech32(msg.Sender)
+	if err != nil {
+		ctx.Logger().Error("invalid sender address", "sender", msg.Sender, "err", err)
+		return nil, types.WrapError(types.ErrInvalidSender, "failed to create cosmos address for sender: %v; error: %v", msg.Sender, err)
+	}
+
+	if err := k.BurnETH(ctx, cosmAddr, msg.Amount); err != nil {
+		ctx.Logger().Error("failed to burn ETH", "cosmosAddress", cosmAddr, "evmAddress", msg.Target, "err", err)
```
joshklop, line 122: pedantic nit: better to capitalize logs.

joshklop (approved):
I had to look pretty hard to find something to comment on haha

LGTM. Thanks!

x/rollup/keeper/msg_server.go
natebeauregard: Sounds good, I'll update this to use the bank keeper. I'm going to make a TODO for updating the deposit msg server to go through the bank keeper as well for minting the L2 ETH. The only reason I ended up using the mint keeper for burning was because there was already a precedent for minting from there. It should make test setup a lot easier too since we'll now only need to mock one keeper

x/rollup/keeper/msg_server.go
natebeauregard: I'd never seen a pointer to the cosmos sdk context before so I was curious if something similar was used in the cosmos sdk itself or any prominent cosmos repos. I couldn't find any references so I figured it's best to just play it safe and pass the context struct around instead of a pointer

x/rollup/keeper/msg_server.go
natebeauregard: Forgot I was already using bank keeper here, that was a last minute change before I made a PR :sweat_smile: I still think it's worth looking into minting from the bank keeper too, I can make a task for that

PR #90, opened 2024-07-31
Documentation Updates

phasewalk1:
PR addresses two points of feedback received on the docsite:

1. Remove instructions to tweak the `genesis.json` file generated with `rolld init`. When following the tutorial in order, `x/testmodule` will be included in the genesis file automatically, and since we instruct the user to pass `--chain-id=1` to the init call, we can leave the genesis file as is.
2. Fix the _edit this page_ links in `docusaurus.config.ts`.

Tagging for review: @joshklop @NiloCK 

## Summary by CodeRabbit

- **Documentation**
	- Simplified the tutorial on creating an application with Monomer by removing detailed configurations for `genesis.json`, which may impact user guidance.
	- Updated the "edit this page" links in the documentation to direct users to the correct repository for the Polymer DAO project.
	- Disabled the blog feature in the documentation configuration, streamlining user access to content.


doc/docusaurus.config.ts
```diff
@@ -38,14 +38,14 @@ const config: Config = {
...
-            'https://github.com/facebook/docusaurus/tree/main/packages/create-docusaurus/templates/shared/',
+            'https://github.com/polymerdao/monomer/tree/main/doc/',
         },
         blog: {
           showReadingTime: true,
           // Please change this to your repo.
           // Remove this to remove the "edit this page" links.
           editUrl:
-            'https://github.com/facebook/docusaurus/tree/main/packages/create-docusaurus/templates/shared/',
+            'https://github.com/polymerdao/monomer/tree/main/doc/',
```
NiloCK, line 48: LGTM, although I didn't run locally to click generated links.

I guess that the `blog` key here isn't actively used, but good to have it set in the right direction anyhow.

phasewalk1: I ran it locally to double check, links are good.

> I guess that the `blog` key here isn't actively used, but good to have it set in the right direction anyhow.

This actually brings up a good point. We can disable the blog feature as a whole if we don't need it.

NiloCK (approved):
Approve on the link fix.

doc/docusaurus.config.ts
joshklop: Sounds good

joshklop (approved):

doc/docusaurus.config.ts
phasewalk1: Cool, I'll remove the blog and then this PR should be good to merge

PR #91, opened 2024-07-31
Update withdrawals storage in monomer EVM

natebeauregard:
- After an application block is finalized, all withdrawal transactions are also passed through to the `L2ToL1MessagePasser` contract in the Monomer EVM. The hash of each withdrawal tx is stored in the contract and is used to prove to L1 that the withdrawal took place on L2.
- Unit tests were added to verify the correct storage behavior of the monomer EVM

## Summary by CodeRabbit

- **New Features**
	- Introduced functionality for initiating withdrawals and checking sent messages between Layer 2 and Layer 1.
	- Added a new `MonomerContractExecuter` for simplified smart contract interactions.
	- Created a `PredeployContracts` function to enhance contract deployment into Ethereum state.

- **Bug Fixes**
	- Improved handling and validation of transaction results during withdrawal processing.

- **Refactor**
	- Streamlined code structure and naming conventions across several modules for better clarity and maintainability.

- **Tests**
	- Implemented unit tests for new functionalities, ensuring correct behavior of state management and message passing.

### DIFF
--- a/contracts/predeploys.go
+++ b/contracts/predeploys.go
@@ -1,25 +1,35 @@
 package contracts
 
 import (
-	"github.com/ethereum-optimism/optimism/op-bindings/predeploys"
+	opcontracts "github.com/ethereum-optimism/optimism/op-bindings/predeploys"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/polymerdao/monomer/bindings/generated"
 )
 
 type Predeploy struct {
-	Address          common.Address
-	DeployedBytecode []byte
+	address          common.Address
+	deployedBytecode []byte
 }
 
 var L2ApplicationStateRootProviderAddr = common.HexToAddress("0x4300000000000000000000000000000000000001")
 
-var Predeploys = []*Predeploy{
+var predeploys = []*Predeploy{
 	{
-		Address:          L2ApplicationStateRootProviderAddr,
-		DeployedBytecode: common.FromHex(bindings.L2ApplicationStateRootProviderMetaData.Bin),
+		address:          L2ApplicationStateRootProviderAddr,
+		deployedBytecode: common.FromHex(bindings.L2ApplicationStateRootProviderMetaData.Bin),
 	},
 	{
-		Address:          predeploys.L2ToL1MessagePasserAddr,
-		DeployedBytecode: common.FromHex(bindings.L2ToL1MessagePasserMetaData.Bin),
+		address:          opcontracts.L2ToL1MessagePasserAddr,
+		deployedBytecode: common.FromHex(bindings.L2ToL1MessagePasserMetaData.Bin),
 	},
+}
+
+func PredeployContracts(ethState *state.StateDB) *state.StateDB {
+	// TODO: investigate using the foundry deploy system for setting up the eth genesis state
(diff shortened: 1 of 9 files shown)


contracts/predeploys.go
```diff
@@ -1,25 +1,35 @@
...
 	},
 	{
-		Address:          predeploys.L2ToL1MessagePasserAddr,
-		DeployedBytecode: common.FromHex(bindings.L2ToL1MessagePasserMetaData.Bin),
+		address:          opcontracts.L2ToL1MessagePasserAddr,
+		deployedBytecode: common.FromHex(bindings.L2ToL1MessagePasserMetaData.Bin),
 	},
 }
+
+func PredeployContracts(ethState *state.StateDB) *state.StateDB {
```
joshklop, line 28: nit: go style avoid stutters `contracts.Predeploy` vs `contracts.PredeployContracts`
```suggestion
func Predeploy(ethState *state.StateDB) *state.StateDB {
```

evm/executer.go
```diff
@@ -0,0 +1,64 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("call contract %v: %v", e.contractAddr, err)
+	}
+	return res, nil
+}
+
+func (params *CallParams) getCallParams() (common.Address, *uint256.Int, []byte) {
+	if params.Sender == nil {
+		params.Sender = &MonomerEVMTxOriginAddress
```
joshklop, line 55: does the sender affect the withdrawal hash?

joshklop (approved):

evm/executer.go
natebeauregard: Yeah it will, the [withdrawal hash](https://github.com/ethereum-optimism/optimism/blob/62c7f3b05a70027b30054d4c8974f44000606fb7/packages/contracts-bedrock/contracts/L2/L2ToL1MessagePasser.sol#L103-L112) is the keccak256 hash of the ABI encoded `WithdrawalTransaction` Solidity struct. The `WithdrawalTransaction` struct contains the tx nonce, sender, target, value, gasLimit, and data.

joshklop: Can we set it to the Cosmos address that sent the original withdrawal tx on the Cosmos side? Maybe we should have address adapters too. We already have `EvmToCosmos` I think.

joshklop: I'm not sure if this causes any problems. I can't think of a case where it does. Does it affect who can withdraw on the L1 side? If not, we can add this to the task board for a later increment.

natebeauregard: I think we should already be doing that for the `InitiateWithdrawal` tx. The sender is [pulled directly](https://github.com/polymerdao/monomer/pull/91/files#diff-d371e466d16c4d92dff29b2806e847c619149cd6de147f1cb9ae7e9511d5da7fR273) off of the cosmos message and then is also used as `msg.sender` [in the monomer EVM](https://github.com/polymerdao/monomer/pull/91/files#diff-8cc76b058dc98fb1dcad70a01907c83f6304526cb4b1388d5cdfa0c5083b7b43R47). The cosmos address (`sdk.AccAddress`) can be ported directly to the eth address (`common.Address`) since they're both represented as `[20]byte`.

joshklop: Ah my bad, you're right. I missed that call somehow. Thanks

PR #93, opened 2024-08-01
Ensure tx was successful before storing withdrawal in EVM

natebeauregard:
Makes sure that the tx was successful before storing the withdrawal hash in the monomer EVM. This will prevent transactions that never burn their L2 ETH from being able to claim their ETH on L1.

## Summary by CodeRabbit

- **New Features**
	- Improved handling of withdrawal messages by ensuring they are processed only when transactions are successful, enhancing efficiency and reducing potential errors.
- **Bug Fixes**
	- Fixed logic to prevent unnecessary processing of withdrawal messages for failed transactions.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -190,16 +190,18 @@ func (b *Builder) Build(ctx context.Context, payload *Payload) (*monomer.Block, 
 			return nil, fmt.Errorf("unmarshal cosmos tx: %v", err)
 		}
 
-		// Check for withdrawal messages.
-		for _, msg := range cosmosTx.GetBody().GetMessages() {
-			withdrawalMsg := new(rollupv1.InitiateWithdrawalRequest)
-			if msg.TypeUrl == cdctypes.MsgTypeURL(withdrawalMsg) {
-				if err := withdrawalMsg.Unmarshal(msg.GetValue()); err != nil {
-					return nil, fmt.Errorf("unmarshal InitiateWithdrawalRequest: %v", err)
-				}
-				// Store the withdrawal message hash in the monomer EVM state db.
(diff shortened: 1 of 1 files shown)


builder/builder.go
```diff
@@ -190,16 +190,18 @@ func (b *Builder) Build(ctx context.Context, payload *Payload) (*monomer.Block,
...
-			withdrawalMsg := new(rollupv1.InitiateWithdrawalRequest)
-			if msg.TypeUrl == cdctypes.MsgTypeURL(withdrawalMsg) {
-				if err := withdrawalMsg.Unmarshal(msg.GetValue()); err != nil {
-					return nil, fmt.Errorf("unmarshal InitiateWithdrawalRequest: %v", err)
-				}
-				// Store the withdrawal message hash in the monomer EVM state db.
-				if err := b.storeWithdrawalMsgInEVM(withdrawalMsg, ethState, header); err != nil {
-					return nil, fmt.Errorf("store withdrawal msg in EVM: %v", err)
+		// Check for withdrawal messages if the tx was successful.
+		if execTxResult.IsOK() {
```
joshklop, line 194: Is it possible for this be false even if the eth is burned?

NiloCK, line 194: Looks reasonable to me.

For understanding. This closes a money printing attack vector where someone submits a tx that:
- contains withdrawal messages
- fails

This could queue the withdrawal to be available to them on L1 despite not executing on L2.

natebeauregard: Yep exactly. They could essentially double their ETH for every attempted withdrawal, however this small check to ensure that the tx was successful should prevent that

NiloCK (approved):

builder/builder.go
joshklop: Beautiful.

PR #94, opened 2024-08-01
coverage: 95.3% of statements

AnkushinDaniil:
can't cover panics in hashData and hashDataAsBinary

### DIFF
--- a/monomer_test.go
+++ b/monomer_test.go
@@ -1,19 +1,189 @@
 package monomer_test
 
 import (
+	"math/big"
+	"strconv"
 	"testing"
+	"time"
 
+	"github.com/cometbft/cometbft/types"
+	opeth "github.com/ethereum-optimism/optimism/op-service/eth"
+	"github.com/ethereum/go-ethereum/beacon/engine"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/polymerdao/monomer"
 	"github.com/polymerdao/monomer/testutils"
 	"github.com/stretchr/testify/require"
+	"github.com/test-go/testify/assert"
 )
 
-func TestBlockToEth(t *testing.T) {
-	l1InfoTx, depositTx, cosmosEthTx := testutils.GenerateEthTxs(t)
-	block := testutils.GenerateBlockFromEthTxs(t, l1InfoTx, []*ethtypes.Transaction{depositTx}, []*ethtypes.Transaction{cosmosEthTx})
-	ethBlock, err := block.ToEth()
-	require.NoError(t, err)
-	for i, ethTx := range []*ethtypes.Transaction{l1InfoTx, depositTx, cosmosEthTx} {
-		require.EqualExportedValues(t, ethTx, ethBlock.Body().Transactions[i])
+func TestChainID(t *testing.T) {
+	n := int64(12345)
+	id := monomer.ChainID(n)
+
+	idStr := strconv.FormatInt(n, 10)
+	assert.Equal(t, idStr, id.String())
+
+	bigInt := big.NewInt(n)
+	assert.Equal(t, bigInt, id.Big())
+
+	hexID := (*hexutil.Big)(bigInt)
+	assert.Equal(t, hexID, id.HexBig())
+}
+
+func TestHeader(t *testing.T) {
+	timestamp := time.Now()
+	header := &monomer.Header{
+		ChainID:    monomer.ChainID(12345),
+		Height:     int64(67890),
+		Time:       uint64(timestamp.Unix()),
+		StateRoot:  common.HexToHash("0x1"),
+		ParentHash: common.HexToHash("0x2"),
+		GasLimit:   uint64(3000000),
+		Hash:       common.HexToHash("0x3"),
 	}
+
+	t.Run("ToComet", func(t *testing.T) {
+		cometHeader := header.ToComet()
+		assert.Equal(t, header.ChainID.String(), cometHeader.ChainID)
+		assert.Equal(t, header.Height, cometHeader.Height)
+		assert.Equal(t, timestamp.Unix(), cometHeader.Time.Unix())
+		assert.Equal(t, header.StateRoot.Bytes(), cometHeader.AppHash.Bytes())
+	})
+
+	t.Run("ToEth", func(t *testing.T) {
+		ethHeader := header.ToEth()
+		assert.Equal(t, header.Height, ethHeader.Number.Int64())
+		assert.Equal(t, uint64(timestamp.Unix()), ethHeader.Time)
+		assert.Equal(t, header.ParentHash, ethHeader.ParentHash)
+		assert.Equal(t, header.StateRoot, ethHeader.Root)
+		assert.Equal(t, header.GasLimit, ethHeader.GasLimit)
+	})
+}
+
+func assertPanic(t *testing.T, f func()) {
+	defer func() {
+		if r := recover(); r == nil {
+			t.Errorf("The code did not panic")
+		}
+	}()
+	f()
+}
+
+func TestBlock(t *testing.T) {
+	header := &monomer.Header{
+		ChainID:    12345,
+		Height:     67890,
+		Time:       uint64(time.Now().Unix()),
+		StateRoot:  common.HexToHash("0x1"),
+		ParentHash: common.HexToHash("0x2"),
+		GasLimit:   3000000,
+		Hash:       common.HexToHash("0x3"),
+	}
+
+	l1InfoTx, depositTx, cosmosEthTx := testutils.GenerateEthTxs(t)
+	txs := ethtypes.Transactions{l1InfoTx, depositTx, cosmosEthTx}
+	block := testutils.GenerateBlockFromEthTxs(t,
+		l1InfoTx,
+		[]*ethtypes.Transaction{depositTx},
+		[]*ethtypes.Transaction{cosmosEthTx},
+	)
+
+	emptyTxs := types.Txs{}
+
+	t.Run("NewBlock", func(t *testing.T) {
+		newBlock := monomer.NewBlock(header, emptyTxs)
+
+		assert.Equal(t, header, newBlock.Header)
+		assert.Equal(t, emptyTxs, newBlock.Txs)
+
+		assertPanic(t, func() { monomer.NewBlock(nil, emptyTxs) })
+		assertPanic(t, func() { monomer.NewBlock(header, nil) })
+		assertPanic(t, func() { monomer.NewBlock(nil, nil) })
+	})
+
+	t.Run("SetHeader", func(t *testing.T) {
+		newBlock, err := monomer.SetHeader(monomer.NewBlock(header, emptyTxs))
+		assert.NoError(t, err)
+		assert.NotEqual(t, common.Hash{}, newBlock.Header.Hash)
+
+		_, err = monomer.SetHeader(nil)
+		assert.Error(t, err)
+	})
+
+	t.Run("MakeBlock", func(t *testing.T) {
+		block, err := monomer.MakeBlock(header, emptyTxs)
+		assert.NoError(t, err)
+		assert.NotEqual(t, common.Hash{}, block.Header.Hash)
+	})
+
+	t.Run("ToEth", func(t *testing.T) {
+		ethBlock, err := block.ToEth()
+		require.NoError(t, err)
+		for i, ethTx := range txs {
+			require.EqualExportedValues(t, ethTx, ethBlock.Body().Transactions[i])
+		}
+
+		newBlock := monomer.NewBlock(header, types.Txs{[]byte("transaction1"), []byte("transaction2")})
+		_, err = newBlock.ToEth()
+		assert.Error(t, err)
+
+		newBlock = nil
+		_, err = newBlock.ToEth()
+		assert.Error(t, err)
+	})
+
+	t.Run("ToCometLikeBlock", func(t *testing.T) {
+		cometLikeBlock := block.ToCometLikeBlock()
+		assert.Equal(t, "0", cometLikeBlock.Header.ChainID)
+		assert.Equal(t, time.Date(1970, time.January, 1, 3, 0, 0, 0, time.Local), cometLikeBlock.Header.Time)
+		assert.Equal(t, int64(0), cometLikeBlock.Header.Height)
+		assert.Equal(t, "0000000000000000000000000000000000000000000000000000000000000000", cometLikeBlock.Header.AppHash.String())
+	})
+}
+
+func TestPayloadAttributes(t *testing.T) {
+	t.Run("ID NoTxPool=false", func(t *testing.T) {
+		pa := &monomer.PayloadAttributes{
+			Timestamp:             uint64(time.Now().Unix()),
+			PrevRandao:            [32]byte{1, 2, 3, 4},
+			SuggestedFeeRecipient: common.HexToAddress("0x5"),
+			NoTxPool:              false,
+			GasLimit:              3000000,
+			ParentHash:            common.HexToHash("0x2"),
+			CosmosTxs:             types.Txs{[]byte("transaction1"), []byte("transaction2")},
+		}
+
+		id := pa.ID()
+		assert.NotNil(t, id)
+
+		newID := pa.ID()
+		assert.Equal(t, id, newID)
+	})
+
+	t.Run("ID NoTxPool=true", func(t *testing.T) {
+		pa := &monomer.PayloadAttributes{
+			Timestamp:             uint64(time.Now().Unix()),
+			PrevRandao:            [32]byte{1, 2, 3, 4},
+			SuggestedFeeRecipient: common.HexToAddress("0x5"),
+			NoTxPool:              true,
+			GasLimit:              3000000,
+			ParentHash:            common.HexToHash("0x2"),
+			CosmosTxs:             types.Txs{[]byte("transaction1"), []byte("transaction2")},
+		}
+
+		id := pa.ID()
+		assert.NotNil(t, id)
+	})
+
+	t.Run("ValidForkchoiceUpdateResult", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


AnkushinDaniil:
@joshklop 
review request

monomer_test.go
```diff
@@ -1,19 +1,190 @@
...
-	ethBlock, err := block.ToEth()
-	require.NoError(t, err)
-	for i, ethTx := range []*ethtypes.Transaction{l1InfoTx, depositTx, cosmosEthTx} {
-		require.EqualExportedValues(t, ethTx, ethBlock.Body().Transactions[i])
+func TestChainID(t *testing.T) {
+	n := int64(12345)
+	id := monomer.ChainID(n)
+
+	idStr := strconv.FormatInt(n, 10)
+	assert.Equal(t, idStr, id.String())
```
joshklop, line 26: prefer `require` to fail the test if the condition isn't true

monomer_test.go
```diff
@@ -1,19 +1,190 @@
...
+		GasLimit:   3000000,
+		Hash:       common.HexToHash("0x3"),
 	}
 }
+
+func TestHeader(t *testing.T) {
+	timestamp := time.Now()
+	header := newTestHeader()
+
+	t.Run("ToComet", func(t *testing.T) {
```
joshklop, line 51: Let's have separate tests for these cases. That way we don't share the `header` and `timestamp` vars

monomer_test.go
```diff
@@ -1,19 +1,190 @@
...
+		ethHeader := header.ToEth()
+		assert.Equal(t, header.Height, ethHeader.Number.Int64())
+		assert.Equal(t, uint64(timestamp.Unix()), ethHeader.Time)
+		assert.Equal(t, header.ParentHash, ethHeader.ParentHash)
+		assert.Equal(t, header.StateRoot, ethHeader.Root)
+		assert.Equal(t, header.GasLimit, ethHeader.GasLimit)
+	})
+}
+
+func assertPanic(t *testing.T, f func()) {
```
joshklop, line 69: I think there is `require.Panics` and `require.NotPanics`

joshklop (changes requested):

monomer_test.go
```diff
@@ -1,19 +1,176 @@
...
+	id := monomer.ChainID(n)
+
+	idStr := strconv.FormatInt(n, 10)
+	require.Equal(t, idStr, id.String())
+
+	bigInt := big.NewInt(n)
+	require.Equal(t, bigInt, id.Big())
+
+	hexID := (*hexutil.Big)(bigInt)
+	require.Equal(t, hexID, id.HexBig())
```
joshklop, line 31: The expected value should come first. `require.Equal(t, id.String(), idStr)`

monomer_test.go
```diff
@@ -1,19 +1,176 @@
...
-	ethBlock, err := block.ToEth()
-	require.NoError(t, err)
-	for i, ethTx := range []*ethtypes.Transaction{l1InfoTx, depositTx, cosmosEthTx} {
-		require.EqualExportedValues(t, ethTx, ethBlock.Body().Transactions[i])
+func TestChainID(t *testing.T) {
+	n := int64(12345)
+	id := monomer.ChainID(n)
+
+	idStr := strconv.FormatInt(n, 10)
+	require.Equal(t, idStr, id.String())
```
joshklop, line 25: try to avoid unnecessary variables
```suggestion
	require.Equal(t, strconv.FormatInt(n, 10), id.String())
```

monomer_test.go
```diff
@@ -1,19 +1,176 @@
...
+		StateRoot:  common.HexToHash("0x1"),
+		ParentHash: common.HexToHash("0x2"),
+		GasLimit:   3000000,
+		Hash:       common.HexToHash("0x3"),
 	}
 }
+
+func TestToComet(t *testing.T) {
+	timestamp := time.Now()
+	header := newTestHeader(uint64(timestamp.Unix()))
```
joshklop, line 48: Can the `newTestHeader` have the timestamp on it? What was the reason for that change?

monomer_test.go
```diff
@@ -1,19 +1,176 @@
...
+	cometHeader := header.ToComet()
+
+	require.Equal(t, header.ChainID.String(), cometHeader.ChainID)
+	require.Equal(t, header.Height, cometHeader.Height)
+	require.Equal(t, timestamp.Unix(), cometHeader.Time.Unix())
+	require.Equal(t, header.StateRoot.Bytes(), cometHeader.AppHash.Bytes())
```
joshklop, line 54: It is better to require equality on the entire struct. This way when fields are added, the test will fail, warning us of a potential issue. Something like:
```suggestion
	require.Equal(t, &cometbft.Header{
		ChainID: header.ChainID.String(),
		Height: header.Height,
		// the rest of the fields.
	}, cometHeader)
```

monomer_test.go
```diff
@@ -1,19 +1,176 @@
...
+func TestToEth(t *testing.T) {
+	timestamp := time.Now()
+	header := newTestHeader(uint64(timestamp.Unix()))
+	ethHeader := header.ToEth()
+
+	require.Equal(t, header.Height, ethHeader.Number.Int64())
+	require.Equal(t, uint64(timestamp.Unix()), ethHeader.Time)
+	require.Equal(t, header.ParentHash, ethHeader.ParentHash)
+	require.Equal(t, header.StateRoot, ethHeader.Root)
+	require.Equal(t, header.GasLimit, ethHeader.GasLimit)
```
joshklop, line 66: Same as above, better to test equality on the entire struct.

monomer_test.go
```diff
@@ -1,19 +1,176 @@
...
+		[]*ethtypes.Transaction{cosmosEthTx},
+	)
+
+	emptyTxs := types.Txs{}
+
+	t.Run("NewBlock", func(t *testing.T) {
+		newBlock := monomer.NewBlock(header, emptyTxs)
+
+		require.Equal(t, header, newBlock.Header)
+		require.Equal(t, emptyTxs, newBlock.Txs)
```
joshklop, line 87: Same as above and in other places: better to test equality on the entire struct.

joshklop (changes requested):

monomer_test.go
AnkushinDaniil: As I can see, idStr or strconv.FormatInt(n, 10) is expected

joshklop: Ahh I see, you're right.

monomer_test.go
```diff
@@ -1,19 +1,201 @@
...
+		Time:            header.Time,
+		UncleHash:       ethtypes.EmptyUncleHash,
+		ReceiptHash:     ethtypes.EmptyReceiptsHash,
+		BaseFee:         common.Big0,
+		WithdrawalsHash: &ethtypes.EmptyWithdrawalsHash,
+		Difficulty:      common.Big0,
+	}, ethHeader)
+}
+
+func TestBlock(t *testing.T) {
```
joshklop, line 76: nit: can we use separate functions for the tests instead of sub-tests here? Sub-tests can be harder to follow when a table-test style isn't possible. E.g., `TestNewBlock`, `TestMakeBlock`, etc.

monomer_test.go
```diff
@@ -1,19 +1,201 @@
...
+
+func TestPayloadAttributes(t *testing.T) {
+	t.Run("ID NoTxPool=false", func(t *testing.T) {
+		pa := newPayloadAttributes()
+
+		id := pa.ID()
+		require.NotNil(t, id)
+
+		newID := pa.ID()
+		require.Equal(t, id, newID)
```
joshklop, line 175: this should be tested in all cases.

monomer_test.go
```diff
@@ -1,19 +1,201 @@
...
+	})
+
+	t.Run("ID NoTxPool=true", func(t *testing.T) {
```
joshklop, line 178: is no tx pool true here?

monomer_test.go
```diff
@@ -1,19 +1,201 @@
...
+	})
+
+	t.Run("ValidForkchoiceUpdateResult", func(t *testing.T) {
+		headBlockHash := common.HexToHash("0x1")
+		payloadID := &engine.PayloadID{}
+
+		result := monomer.ValidForkchoiceUpdateResult(&headBlockHash, payloadID)
+		require.Equal(t, opeth.ExecutionValid, result.PayloadStatus.Status)
+		require.Equal(t, headBlockHash.Bytes(), result.PayloadStatus.LatestValidHash.Bytes())
+		require.Equal(t, payloadID, result.PayloadID)
```
joshklop, line 192: do we need these anymore?

joshklop (changes requested):

monomer_test.go
```diff
@@ -1,19 +1,211 @@
...
+		trie.NewStackTrie(nil),
+	), ethBlock)
+
+	newBlock := monomer.NewBlock(newTestHeader(), bfttypes.Txs{[]byte("transaction1"), []byte("transaction2")})
+	_, err = newBlock.ToEth()
+	require.Error(t, err)
+
+	newBlock = nil
+	_, err = newBlock.ToEth()
+	require.Error(t, err)
```
joshklop, line 145: Are there any additional tests done here, or is it mainly refactoring?

monomer_test.go
```diff
@@ -1,19 +1,211 @@
...
+	)
+	cometLikeBlock := block.ToCometLikeBlock()
+	require.Equal(t, &bfttypes.Block{
+		Header: bfttypes.Header{
+			ChainID: block.Header.ChainID.String(),
+			Time:    time.Unix(int64(block.Header.Time), 0),
+			Height:  block.Header.Height,
+			AppHash: block.Header.StateRoot.Bytes(),
+		},
+	}, cometLikeBlock)
```
joshklop, line 163: variable is used only once

monomer_test.go
```diff
@@ -1,19 +1,211 @@
...
 }
+
+func TestPayloadAttributesIDNoTxPoolIsFalse(t *testing.T) {
+	pa := newPayloadAttributes()
+
+	id := pa.ID()
+	require.NotNil(t, id)
+
+	newID := pa.ID()
+	require.Equal(t, id, newID)
```
joshklop, line 185: variable is used only once

joshklop (changes requested):

monomer_test.go
AnkushinDaniil: There are additional error case tests

monomer_test.go
```diff
@@ -1,19 +1,213 @@
...
+		trie.NewStackTrie(nil),
+	), ethBlock)
+
+	newBlock := monomer.NewBlock(newTestHeader(), bfttypes.Txs{[]byte("transaction1"), []byte("transaction2")})
+	_, err = newBlock.ToEth()
+	require.Error(t, err)
+
+	newBlock = nil
+	_, err = newBlock.ToEth()
+	require.Error(t, err)
```
joshklop, line 145: We don't need to test this case imo. It's generally assumed the struct must be non-nil before calling its methods.

monomer_test.go
```diff
@@ -1,19 +1,213 @@
...
+		ethTxs,
+		nil,
+		[]*ethtypes.Receipt{},
+		[]*ethtypes.Withdrawal{},
+		trie.NewStackTrie(nil),
+	), ethBlock)
+
+	newBlock := monomer.NewBlock(newTestHeader(), bfttypes.Txs{[]byte("transaction1"), []byte("transaction2")})
+	_, err = newBlock.ToEth()
+	require.Error(t, err)
```
joshklop, line 141: can we add a comment describing what this case is testing? It may be even better to separate this out into a separate test like `TestBlockToEthFailsWithoutDepositTxs` (long names are ok in tests; the test name should generally describe what the test does so we don't need comments)

monomer_test.go
```diff
@@ -1,19 +1,213 @@
...
 	require.NoError(t, err)
-	for i, ethTx := range []*ethtypes.Transaction{l1InfoTx, depositTx, cosmosEthTx} {
-		require.EqualExportedValues(t, ethTx, ethBlock.Body().Transactions[i])
+	require.EqualExportedValues(t, ethtypes.NewBlockWithWithdrawals(
+		block.Header.ToEth(),
+		ethTxs,
+		nil,
+		[]*ethtypes.Receipt{},
+		[]*ethtypes.Withdrawal{},
+		trie.NewStackTrie(nil),
```
joshklop, line 136: We're getting pretty close to testing the implementation here. Let's expect the final result, rather than passing to a helper. Something like:

```golang
require.EqualExportedValues(t, &ethtypes.Block{
	Header: block.Header.ToEth(),
	// ... the other fields ...
}, ethBlock)
```

The test for `ToCometLikeBlock` is good. We should mimic that here.

monomer_test.go
```diff
@@ -1,19 +1,213 @@
...
+func TestPayloadAttributesIDNoTxPoolIsTrue(t *testing.T) {
+	pa := newPayloadAttributes()
+	pa.NoTxPool = true
+
+	id := pa.ID()
+	require.NotNil(t, id)
+
+	newID := pa.ID()
+	require.Equal(t, id, newID)
+}
```
joshklop, line 199: I didn't notice this before, but I think we can turn this into a table test.

joshklop (changes requested):
Getting close! Nice work.

monomer_test.go
AnkushinDaniil: ethtypes.Block's fields are private:
```go
type Block struct {
    header       *Header
    uncles       []*Header
    transactions Transactions
    withdrawals  Withdrawals

    // caches
    hash atomic.Value
    size atomic.Value

    // These fields are used by package eth to track
    // inter-peer block relay.
    ReceivedAt   time.Time
    ReceivedFrom interface{}
}
```
When I update 
```suggestion
	require.Equal(t, ethtypes.NewBlockWithWithdrawals(
		block.Header.ToEth(),
		ethTxs,
		nil,
		[]*ethtypes.Receipt{},
		[]*ethtypes.Withdrawal{},
		trie.NewStackTrie(nil),
```
I have an error
```
Error Trace:    /Users/daniilankusin/monomer/monomer_test.go:130
                Error:          Not equal: 
                                expected: &types.Block{header:(*types.Header)(0x140004e2508), uncles:[]*types.Header(nil), transactions:types.Transactions{(*types.Transaction)(0x140003bc070), (*types.Transaction)(0x140003bc150), (*types.Transaction)(0x140003bc1c0)}, withdrawals:types.Withdrawals{}, hash:atomic.Value{v:interface {}(nil)}, size:atomic.Value{v:interface {}(nil)}, ReceivedAt:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), ReceivedFrom:interface {}(nil)}
                                actual  : &types.Block{header:(*types.Header)(0x140004e2008), uncles:[]*types.Header(nil), transactions:types.Transactions{(*types.Transaction)(0x140003bc310), (*types.Transaction)(0x140003bc380), (*types.Transaction)(0x140003bc3f0)}, withdrawals:types.Withdrawals{}, hash:atomic.Value{v:interface {}(nil)}, size:atomic.Value{v:interface {}(nil)}, ReceivedAt:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), ReceivedFrom:interface {}(nil)}
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -122,4 +122,4 @@
                                    time: (time.Time) {
                                -    wall: (uint64) 13953742264056089080,
                                -    ext: (int64) 50291918,
                                +    wall: (uint64) 13953742264056103080,
                                +    ext: (int64) 50306001,
                                     loc: (*time.Location)({
                                @@ -955,4 +955,4 @@
                                    time: (time.Time) {
                                -    wall: (uint64) 13953742264056096080,
                                -    ext: (int64) 50299209,
                                +    wall: (uint64) 13953742264056104080,
                                +    ext: (int64) 50307084,
                                     loc: (*time.Location)({
                                @@ -1579,4 +1579,4 @@
                                    time: (time.Time) {
                                -    wall: (uint64) 13953742264056098080,
                                -    ext: (int64) 50301126,
                                +    wall: (uint64) 13953742264056105080,
                                +    ext: (int64) 50307751,
                                     loc: (*time.Location)({
                Test:           TestBlockToEth
```
It's impossible to change the time here, so I'm going to rewrite test to handle fields separately.

monomer_test.go
AnkushinDaniil: But we have this if statement in the `ToEth` method
```go
func (b *Block) ToEth() (*ethtypes.Block, error) {
	if b == nil {
		return nil, errors.New("converted a nil block")
	}
	// code
}
```

joshklop: Ah, that's new, thanks for pointing that out. We can remove that check. To maintain that convention, we would need to spread such checks everywhere.

AnkushinDaniil: fixed

monomer_test.go
```diff
@@ -1,19 +1,248 @@
...
+	require.Equal(t, expectedEthBlock.ReceiptHash(), actualEthBlock.ReceiptHash())
+	require.Equal(t, expectedEthBlock.Root(), actualEthBlock.Root())
+	require.Equal(t, expectedEthBlock.SanityCheck(), actualEthBlock.SanityCheck())
+	require.Equal(t, expectedEthBlock.Size(), actualEthBlock.Size())
+	require.Equal(t, expectedEthBlock.Time(), actualEthBlock.Time())
+	require.Equal(t, expectedEthBlock.Transactions().Len(), actualEthBlock.Transactions().Len())
+	require.Equal(t, expectedEthBlock.TxHash(), actualEthBlock.TxHash())
+	require.Equal(t, expectedEthBlock.UncleHash(), actualEthBlock.UncleHash())
+	require.Equal(t, expectedEthBlock.Uncles(), actualEthBlock.Uncles())
+	require.Equal(t, expectedEthBlock.Withdrawals(), actualEthBlock.Withdrawals())
```
joshklop, line 162: Actually, I just thought of this... what if we just compare the hashes? Should be equivalent to comparing everything else...
```suggestion
	require.Equal(t, expectedEthBlock.Hash(), actualEthBlock.Hash())
```

joshklop (approved):
I think there's just this last thing to fix. Please ping me when it's ready to merge. Great work

joshklop (approved):

PR #95, opened 2024-08-02
added error handler in engine api if block build fails

AnkushinDaniil:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling in the block commitment process to provide clearer error messages while maintaining the panic behavior, enhancing debugging capabilities.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -250,7 +250,7 @@ func (e *EngineAPI) GetPayloadV3(ctx context.Context, payloadID engine.PayloadID
 		NoTxPool:             e.currentPayloadAttributes.NoTxPool,
 	})
 	if err != nil {
-		return nil, engine.GenericServerError.With(fmt.Errorf("build block: %v", err))
+		panic(fmt.Errorf("build block: %v", err))
 	}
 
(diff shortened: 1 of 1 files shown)


AnkushinDaniil:
@joshklop 
review request

engine/engine.go
```diff
@@ -251,7 +250,7 @@ func (e *EngineAPI) GetPayloadV3(ctx context.Context, payloadID engine.PayloadID
 		NoTxPool:             e.currentPayloadAttributes.NoTxPool,
 	})
 	if err != nil {
-		log.Panicf("failed to commit block: %v", err) // TODO error handling. An error here is potentially a big problem.
+		return nil, engine.GenericServerError.With(fmt.Errorf("build block: %v", err))
```
joshklop, line 253: Removing `log` is good, but let's keep the panic here for now. We can't really recover from an error here. We have a few work streams in progress to fix it, but it will take a while (db consistency, etc.).
```suggestion
		return panic(fmt.Errorf("build block: %v", err))
```

joshklop (changes requested):

NiloCK (approved):

PR #98, opened 2024-08-02
Add message nonce to the InitiateWithdrawal event attributes

natebeauregard:
In order to prove that a withdrawal occurred, a relayer will need to include the full withdrawal tx information in their proof. Unfortunately, the message nonce is calculated inside of the `L2ToL1MessagePasser` so it isn't available whenever the `x/rollup` event for initiating the withdrawal is created.

To allow relayers to have the full withdrawal tx, the message nonce is appended onto the tx's event attributes after the withdrawal hash is stored in the EVM. Additionally, some other attributes like `gasLimit` and `data` were added as well to the `InitiateWithdrawal` event in the `x/rollup` module make it easy for the relayer to build their withdrawal proof. 

Unit tests aren't added for the `builder.go` changes but will be added in a subsequent PR once the testapp supports `x/rollup`.

## Summary by CodeRabbit

- **New Features**
  - Added a new method to retrieve the message nonce in the `L2ToL1MessagePasserExecuter`.
  - Enhanced the functionality of withdrawal message handling in the `Builder` class.

- **Improvements**
  - Streamlined naming conventions for constants throughout the system for better clarity and consistency.
  - Updated event logging in the `InitiateWithdrawal` method to include additional attributes.

- **Documentation**
  - Improved comment clarity in the `InitiateWithdrawalRequest` message structure. 

- **Bug Fixes**
  - Adjusted variable names and error handling for improved functionality and maintainability across various functions.


proto/rollup/v1/msg.proto
```diff
@@ -22,15 +22,16 @@ message ApplyL1TxsRequest {
...
+// TODO: add validations in the x/rollup module to ensure that the target address and gas limit are valid before burning the users ETH
+// InitiateWithdrawalRequest defines a message for all L2 withdrawal txs
 message InitiateWithdrawalRequest {
   option (cosmos.msg.v1.signer) = "sender";
   // The cosmos address of the user who wants to withdraw from L2.
   string sender = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
   // The ethereum address on L1 that the user wants to withdraw to.
   string target = 2;
-  // The amount of ETH that the user wants to withdraw.
-  string amount = 3 [
```
natebeauregard, line 33: I renamed `amount` to `value` to have parity with the Ethereum withdrawal tx type.

proto/rollup/v1/msg.proto
```diff
@@ -22,15 +22,16 @@ message ApplyL1TxsRequest {
 
 message ApplyL1TxsResponse {}
 
-// MsgInitiateWithdrawal defines a message for all L2 withdrawal txs
+// TODO: add validations in the x/rollup module to ensure that the target address and gas limit are valid before burning the users ETH
```
natebeauregard, line 25: I created a subtask for this on the withdrawals task. We should validate the target address and gas limit here and ensure that an invalid withdrawals hash can't be submitted to the `L1ToL2MessagePasser` contract.

joshklop (commented):
Looks great on a first read-through! I'll take a closer look later. Quick nit for future reference: usually renames are better to go in a separate commit to keep diffs clean and easily reviewable.

proto/rollup/v1/msg.proto
NiloCK: Clarifying: "Ethereum withdrawal tx type" here refers to the OP stack [withdrawal event type](https://specs.optimism.io/protocol/withdrawals.html#the-l2tol1messagepasser-contract) and [L1 finalizing tx type](https://specs.optimism.io/protocol/withdrawals.html#withdrawal-verification-and-finalization)?

natebeauregard: There is a universal `WithdrawalTransaction` [struct](https://github.com/ethereum-optimism/optimism/blob/08daf8dbd38c9ffdbd18fc9a211c227606cdb0ad/packages/contracts-bedrock/src/libraries/Types.sol#L55-L69) created in OP Stack's `Types.sol` that is used to build the initiating, proving, and finalizing transactions on both L2 and L1. This is the struct that I'm referring to here

builder/builder.go
```diff
@@ -168,43 +170,56 @@ func (b *Builder) Build(ctx context.Context, payload *Payload) (*monomer.Block,
...
+						if event.GetType() == rolluptypes.EventTypeWithdrawalInitiated {
+							event.Attributes = append(event.Attributes, abcitypes.EventAttribute{
+								Key:   rolluptypes.AttributeKeyNonce,
+								Value: hexutil.Encode(nonce.Bytes()),
+							})
+						}
+					}
 				}
 			}
 		}
```
NiloCK, line 213: Builder.Build() is getting long-winded and the involved logic spans some different domains.

The highlighted chunk might be a candidate for encapsulating as a smaller scoped `parseWithdrawal` function.

builder/builder.go
```diff
@@ -261,25 +276,32 @@ func (b *Builder) storeWithdrawalMsgInEVM(
 	withdrawalMsg *rollupv1.InitiateWithdrawalRequest,
 	ethState *state.StateDB,
 	header *monomer.Header,
-) error {
+) (*big.Int, error) {
```
NiloCK, line 279: Should add the meaning of this return value to `storeWithdrawalMsgInEVM`'s godoc.

NiloCK (approved):
Looks great.

For my understanding - I don't have a full picture in my head of how the monomer EVM state updates relate to block production.

`executor.GetMessageNonce` checks the L2 chain state for the global withdrawal nonce, as recorded on [The L2ToL1MessagePasser Contract](https://specs.optimism.io/protocol/withdrawals.html#the-l2tol1messagepasser-contract).

Does the current implementation expect to correctly handle multiple withdrawal initiating transactions in a single block? IE, will the `executor`'s view of the state change mid block here?

natebeauregard:
> Looks great.
> 
> For my understanding - I don't have a full picture in my head of how the monomer EVM state updates relate to block production.
> 
> `executor.GetMessageNonce` checks the L2 chain state for the global withdrawal nonce, as recorded on [The L2ToL1MessagePasser Contract](https://specs.optimism.io/protocol/withdrawals.html#the-l2tol1messagepasser-contract).
> 
> Does the current implementation expect to correctly handle multiple withdrawal initiating transactions in a single block? IE, will the `executor`'s view of the state change mid block here?

Yeah that's a good question. We're [creating](https://github.com/polymerdao/monomer/blob/03efcdf520b3f41eeb737f1f72e4c2e06c50ac93/builder/builder.go#L166) the current ethereum state from the DB at the beginning of the `Build` call before any EVM instructions are executed and then [committing](https://github.com/polymerdao/monomer/blob/03efcdf520b3f41eeb737f1f72e4c2e06c50ac93/builder/builder.go#L210) the updated state after all EVM instructions are executed. With this approach we should be able to handle multiple withdrawal initiating messages in a single block (or even a single tx with multiple withdrawal messages!) since we're executing on top of the eth state created at the beginning of the call.

Unfortunately, we can't test this right now since the builder testapp doesn't support the `x/rollup` module yet.

PR #103, opened 2024-08-06
Add ValidateBasic func

### LINKED ISSUE #102: Add additional InitiateWithdrawalRequest message validations
We should add validations for the `InitiateWithdrawalRequest` in `x/rollup` to ensure that a proper config for proving the withdrawal on L1 was supplied. The `sender` address and `value` are validated in the rollup module, however we should also ensure that the `target` field is a valid Ethereum address and that the `gas_limit` and `data` field are valid. The validations should be done in a `ValidateBasic` [method](https://docs.cosmos.network/main/build/building-modules/messages-and-queries#transactionmsg-interface) on the `InitiateWithdrawalRequest` message type.

For the `target` field validation, it's probably sufficient to use `common.IsHexAddress`. The `gas_limit` and `data` fields may not require any validations, however the `L2ToL1MessagePasser` contract should be checked to be sure.

AnkushinDaniil:
fix #102 

## Summary by CodeRabbit

- **New Features**
	- Introduced a validation process for withdrawal requests, ensuring that key fields such as Ethereum address and gas limit are correctly verified before processing.
- **Bug Fixes**
	- Enhanced error handling for invalid withdrawal requests, providing clearer feedback to users when issues arise.
- **Tests**
	- Added a comprehensive test suite to validate the new withdrawal request validation logic, ensuring reliability and robustness against various input scenarios.

### DIFF
--- a/x/rollup/keeper/msg_server.go
+++ b/x/rollup/keeper/msg_server.go
@@ -233,7 +233,7 @@ func evmToCosmos(addr common.Address) sdk.AccAddress {
 // TODO: This is a temporary change while the rollup module refactor is being done. Change when possible
 func validateBasic(m *rollupv1.InitiateWithdrawalRequest) error {
 	// We should ensure that the target field is a valid Ethereum address and that the gas_limit and data field are valid.
-	if common.IsHexAddress(m.Target) {
+	if !common.IsHexAddress(m.Target) {
 		return fmt.Errorf("invalid Ethereum address: %s", m.Target)
 	}
 	// Ensure gas_limit is within a reasonable range
--- a/x/rollup/keeper/msg_server.go
+++ b/x/rollup/keeper/msg_server.go
@@ -242,5 +242,6 @@ func validateBasic(m *rollupv1.InitiateWithdrawalRequest) error { //nolint:unuse
 		return fmt.Errorf("gas limit must be between 5,000 and 9,223,372,036,854,775,807: %d", gasLimit)
 	}
 	// The Data field should be formatted according to the ABI of the target address.
+	// https://docs.optimism.io/stack/protocol/rollup/withdrawal-flow
 	return nil
 }
--- a/x/rollup/keeper/msg_server.go
+++ b/x/rollup/keeper/msg_server.go
@@ -3,7 +3,6 @@ package keeper
 import (
 	"context"
 	"encoding/binary"
-	"encoding/hex"
 	"encoding/json"
 	"fmt"
 
@@ -242,11 +241,6 @@ func validateBasic(m *rollupv1.InitiateWithdrawalRequest) error { //nolint:unuse
 	if gasLimit < params.MinGasLimit || gasLimit > params.MaxGasLimit {
 		return fmt.Errorf("gas limit must be between 5,000 and 9,223,372,036,854,775,807: %d", gasLimit)
 	}
-	// The Data field should be formatted according to the ABI of the target address.
-	// https://docs.optimism.io/stack/protocol/rollup/withdrawal-flow
-	_, err := hex.Decode([]byte{}, m.Data)
-	if err != nil {
-		return fmt.Errorf("invalid data field: %s", m.Data)
-	}
+
 	return nil
 }
--- a/x/rollup/keeper/msg_server.go
+++ b/x/rollup/keeper/msg_server.go
@@ -13,6 +13,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/params"
 	rollupv1 "github.com/polymerdao/monomer/gen/rollup/v1"
 	"github.com/polymerdao/monomer/x/rollup/types"
 	"github.com/samber/lo"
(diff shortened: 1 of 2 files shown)


x/rollup/keeper/msg_server.go
```diff
@@ -227,3 +229,24 @@
 func evmToCosmos(addr common.Address) sdk.AccAddress {
 	return addr.Bytes()
 }
+
+// TODO: This is a temporary change while the rollup module refactor is being done. Change when possible
+func ValidateBasic(m *rollupv1.InitiateWithdrawalRequest) error {
+	// We should ensure that the target field is a valid Ethereum address and that the gas_limit and data field are valid.
+	if common.IsHexAddress(m.Target) {
+		return fmt.Errorf("invalid Ethereum address: %s", m.Target)
```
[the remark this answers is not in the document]
natebeauregard: I agree - we should check `!common.IsHexAddress` here

x/rollup/keeper/msg_server.go
```diff
@@ -227,3 +229,24 @@ func (k *Keeper) SetL1BlockHistory(ctx context.Context, info *derive.L1BlockInfo
...
+	}
+	// Ensure gas_limit is within a reasonable range
+	// https://insights.deribit.com/industry/ethereums-gas-mechanics/
+	gasLimit := binary.BigEndian.Uint64(m.GasLimit) // size=24 (0x18), offset=64 (0x40)
+	if gasLimit < 21000 || gasLimit > 15000000 {
+		return fmt.Errorf("gas limit must be between 21,000 and 15,000,000: %d", gasLimit)
```
natebeauregard, line 243: Where did this upper bound of `15000000` come from? Are there existing constants or helper funcs from geth that we can reuse here instead of needing to redeclare the gas limit boundaries?

x/rollup/keeper/msg_server.go
```diff
@@ -227,3 +229,24 @@ func (k *Keeper) SetL1BlockHistory(ctx context.Context, info *derive.L1BlockInfo
 func evmToCosmos(addr common.Address) sdk.AccAddress {
 	return addr.Bytes()
 }
+
+// TODO: This is a temporary change while the rollup module refactor is being done. Change when possible
+func ValidateBasic(m *rollupv1.InitiateWithdrawalRequest) error {
```
natebeauregard, line 234: It's probably worth renaming this `validateBasic` for now while the func is declared here since it doesn't need to be exported and to avoid confusion with the message interface. It will be renamed to `ValidateBasic` with the message as the method receiver after the rollup refactor is done.

x/rollup/keeper/msg_server.go
```diff
@@ -227,3 +229,24 @@ func (k *Keeper) SetL1BlockHistory(ctx context.Context, info *derive.L1BlockInfo
...
+		return fmt.Errorf("invalid Ethereum address: %s", m.Target)
+	}
+	// Ensure gas_limit is within a reasonable range
+	// https://insights.deribit.com/industry/ethereums-gas-mechanics/
+	gasLimit := binary.BigEndian.Uint64(m.GasLimit) // size=24 (0x18), offset=64 (0x40)
+	if gasLimit < 21000 || gasLimit > 15000000 {
+		return fmt.Errorf("gas limit must be between 21,000 and 15,000,000: %d", gasLimit)
+	}
+	// The data field should be in valid hex format
+	_, err := hex.Decode([]byte{}, m.Data)
```
natebeauregard, line 246: Were you able to verify that the `L2ToL1MessagePasser` and `OptimismPortal` contracts expect data to always be hex? Could it have a different encoding instead and still work? I'm curious if we need to validate anything here, it's probably worth checking to see how this is handled in Solidity

AnkushinDaniil: As I can see, we don't need to validate date field. 
[Here](https://docs.optimism.io/stack/protocol/rollup/withdrawal-flow) is information about data field in withdrawal flows. 
[Here](https://github.com/ethereum-optimism/optimism/blob/62c7f3b05a70027b30054d4c8974f44000606fb7/packages/contracts-bedrock/contracts/L2/L2ToL1MessagePasser.sol#L98) we have initiateWithdrawal function.
There are no requirements for data field.
If you know more about data flow after this function, please share.

natebeauregard (commented):
Looking really good! Great catch with the gas limit validations. I left a few comments and suggestions

x/rollup/keeper/msg_server.go
```diff
@@ -239,7 +244,7 @@ func ValidateBasic(m *rollupv1.InitiateWithdrawalRequest) error {
 	}
 	// Check if the gas limit is within the allowed range.
 	gasLimit := new(big.Int).SetBytes(m.GasLimit).Uint64() // size=24 (0x18), offset=64 (0x40)
-	if gasLimit < params.TxGas || gasLimit > params.MaxGasLimit {
+	if gasLimit < MinGasLimit || gasLimit > MaxGasLimit {
 		return fmt.Errorf("gas limit must be between %d and %d: %d", params.TxGas, params.MaxGasLimit, gasLimit)
```
natebeauregard, line 248: Can you use the constants instead of the `params.` values in the error message creation as well?

x/rollup/keeper/msg_server.go
```diff
@@ -18,6 +18,11 @@ import (
 	"github.com/samber/lo"
 )
 
+const (
+	MinGasLimit = params.TxGas
+	MaxGasLimit = params.MaxGasLimit
```
natebeauregard, line 23: We should probably clarify that these are for the tx gas limit and not the block gas limit. Could you rename these to something like `MinTxGasLimit`/`MaxTxGasLimit`?

x/rollup/keeper/msg_server.go
```diff
@@ -239,7 +244,7 @@ func ValidateBasic(m *rollupv1.InitiateWithdrawalRequest) error {
 	}
 	// Check if the gas limit is within the allowed range.
 	gasLimit := new(big.Int).SetBytes(m.GasLimit).Uint64() // size=24 (0x18), offset=64 (0x40)
```
natebeauregard, line 246: Does the `// size=24 (0x18), offset=64 (0x40)` comment apply here? Could it be removed?

natebeauregard (approved):
LGTM! Left a couple nits, I'll merge once they're updated

Issue #111, opened 2024-08-07
Return 


joshklop:
Instead of passing in an `ethdb.Database`

https://github.com/polymerdao/monomer/blob/46378dfdb1caf2bae001da22299936c5597bac81/node/node.go#L62

and then promptly converting it to a `state.Database`

https://github.com/polymerdao/monomer/blob/46378dfdb1caf2bae001da22299936c5597bac81/node/node.go#L73

let's just allow the user to pass in a `state.Database` to begin with.

PR #114, opened 2024-08-08
Refactor node.New and ProofProvider.getState

Jesserc:
Accept state.Database in node.New instead of ethdb.Database
- Replace ethdb.Database parameter with state.Database in node.New
- Remove conversion to state.Database inside the function
- Update related function calls and tests

## Summary by CodeRabbit

- **New Features**
	- Enhanced database management for Ethereum state, improving functionality and robustness.

- **Bug Fixes**
	- Improved error handling for closing various database components, ensuring better resource management.

- **Refactor**
	- Updated database initialization and type usage across multiple components for optimized performance.

- **Tests**
	- Enhanced test database handling to align with new database management practices, improving test reliability.


Jesserc:
@joshklop, please let me know if this fixes #111

node/node_test.go
```diff
@@ -34,9 +35,10 @@ func TestRun(t *testing.T) {
 	cometListener, err := net.Listen("tcp", cometHTTPAddress)
 	require.NoError(t, err)
 	app := testapp.NewTest(t, chainID.String())
-	ethstatedb := rawdb.NewMemoryDatabase()
+	ethstatedb := state.NewDatabase(rawdb.NewMemoryDatabase())
 	defer func() {
-		require.NoError(t, ethstatedb.Close())
+		require.NoError(t, ethstatedb.DiskDB().Close())
+		require.NoError(t, ethstatedb.TrieDB().Close())
```
joshklop, line 41: Can we use `testutils.NewEthStateDB` instead?

e2e/stack.go
```diff
@@ -243,8 +243,9 @@ func (s *Stack) runMonomer(ctx context.Context, env *environment.Env, genesisTim
 	env.DeferErr("close tx db", txdb.Close)
 	mempooldb := dbm.NewMemDB()
 	env.DeferErr("close mempool db", mempooldb.Close)
-	ethstatedb := rawdb.NewMemoryDatabase()
-	env.DeferErr("close eth state db", ethstatedb.Close)
+	ethstatedb := state.NewDatabase(rawdb.NewMemoryDatabase())
+	env.DeferErr("close eth state disk db", ethstatedb.DiskDB().Close)
+	env.DeferErr("close eth state trie db", ethstatedb.TrieDB().Close)
```
joshklop, line 248: We should close the rawdb. No need to close individual instances of DBs obtained from the statedb.

something like:
```golang
raw := rawdb.NewMemoryDatabase()
env.DeferErr("close raw db", raw.Close)
ethstatedb := state.NewDatabase(raw)
```

eth/get_proof.go
```diff
@@ -60,22 +60,22 @@ func (p *ProofProvider) getState(blockNumber *big.Int) (*state.StateDB, types.He
 		block, err = p.blockStore.BlockByHeight(blockNumber.Uint64())
 	}
 	if err != nil {
-		return nil, types.Header{}, fmt.Errorf("get eth block %d: %w", blockNumber, err)
+		return nil, &types.Header{}, fmt.Errorf("get eth block %d: %w", blockNumber, err)
```
joshklop, line 63: This will allocate memory unnecessarily. We should return `nil` instead.

eth/get_proof.go
```diff
@@ -60,22 +60,22 @@ func (p *ProofProvider) getState(blockNumber *big.Int) (*state.StateDB, types.He
...
+		return nil, &types.Header{}, fmt.Errorf("convert block to Ethereum representation: %v", err)
 	}
 
 	header := ethBlock.Header()
 	hash := ethBlock.Hash()
 
 	sdb, err := state.New(hash, p.database, nil)
 	if err != nil {
-		return nil, *header, fmt.Errorf("opening state.StateDB: %w", err)
+		return nil, header, fmt.Errorf("opening state.StateDB: %w", err)
```
joshklop, line 75: When returning errors, it's best to return `nil` for all other types.

joshklop (changes requested):

node/node_test.go
```diff
@@ -35,7 +33,7 @@ func TestRun(t *testing.T) {
 	cometListener, err := net.Listen("tcp", cometHTTPAddress)
 	require.NoError(t, err)
 	app := testapp.NewTest(t, chainID.String())
-	ethstatedb := state.NewDatabase(rawdb.NewMemoryDatabase())
+	ethstatedb := testutils.NewEthStateDB(t)
```
joshklop, line 36: Do we need this extra variable? I think we can place the `testutils.NewEthStateDB` call with the other similar calls on line 52.

node/node_test.go
```diff
@@ -34,9 +33,10 @@ func TestRun(t *testing.T) {
 	cometListener, err := net.Listen("tcp", cometHTTPAddress)
 	require.NoError(t, err)
 	app := testapp.NewTest(t, chainID.String())
-	ethstatedb := rawdb.NewMemoryDatabase()
+	ethstatedb := testutils.NewEthStateDB(t)
 	defer func() {
-		require.NoError(t, ethstatedb.Close())
+		require.NoError(t, ethstatedb.DiskDB().Close())
+		require.NoError(t, ethstatedb.TrieDB().Close())
```
joshklop, line 39: Why do we need to close these manually? Is this documented anywhere?

node/node_test.go
Jesserc: Do you mean doing this directly on line 52?
```
n := node.New(
//...other code
testutils.NewEthStateDB(t)
//...other code
)
```

node/node_test.go
joshklop: Looking at this again, I can see why this might be necessary. It still feels confusing that we would need to close them both ourselves (I would expect to use something like `ethstatedb.Close()`). 

Can you find any examples of this pattern in geth itself? I see cases these `Close` methods are not called.

Jesserc: I'm not sure, the original code had this
```
ethstatedb := rawdb.NewMemoryDatabase()
	defer func() {
		require.NoError(t, ethstatedb.Close())
	}()
```
So I just followed the convention

node/node_test.go
joshklop: Yes.

node/node_test.go
joshklop: That's closing the rawdb, which we do in `testutils.NewEthStateDB`. I'm curious if we need to close the dbs returned on the `state.Database` interface.

Jesserc: Here's some example from geth: https://github.com/ethereum/go-ethereum/blob/83e70aa3d00bbee9713b31768a30a7741eee8945/core/state/statedb_test.go#L277 and https://github.com/ethereum/go-ethereum/blob/83e70aa3d00bbee9713b31768a30a7741eee8945/eth/state_accessor.go#L71

The dbs returned by `state.Database` is not closed

joshklop: This one nerd-sniped me. It looks like the correct way to manage and close a `state.Database` is undocumented.

The `DiskDB()` is the underlying database, the `ethdb.Database` in our case. We defer the close when we open it ourselves.

It appears the TrieDB, on the other hand, _does_ need to be closed. Although there are a few places where geth does not close the TrieDB, they are all ephemeral (genesis state handling, for example). The trie db that get maintains on its `BlockChain` object is closed when the blockchain is stopped.

This means our current way of creating a `state.Database` error-prone (it's easy to forget to close the `TrieDB`). As a remedy, we can instead use the `state.NewDatabaseWithNodeDB` function to pass in the triedb separately.

I'll create a separate issue for this. For now @Jesserc, go ahead and remove the line that closes the `DiskDB`, but leave the line that closes the `TrieDB`. This seemingly innocuous change led to some important realizations! Thanks for your work here.

cc @natebeauregard this may be interesting and relevant for you in future withdrawals work.

e2e/stack.go
```diff
@@ -243,8 +243,9 @@ func (s *Stack) runMonomer(ctx context.Context, env *environment.Env, genesisTim
 	env.DeferErr("close tx db", txdb.Close)
 	mempooldb := dbm.NewMemDB()
 	env.DeferErr("close mempool db", mempooldb.Close)
-	ethstatedb := rawdb.NewMemoryDatabase()
-	env.DeferErr("close eth state db", ethstatedb.Close)
+	raw := rawdb.NewMemoryDatabase()
+	env.DeferErr("close raw db", raw.Close)
+	ethstatedb := state.NewDatabase(raw)
```
Jesserc, line 248: following up with the last comment, Should I close the `TrieDB` here too or leave it as it is?

joshklop: Yes, go ahead and close the `TrieDB`. Thanks for pointing this case out!

e2e/stack.go
```diff
@@ -243,8 +243,10 @@ func (s *Stack) runMonomer(ctx context.Context, env *environment.Env, genesisTim
...
+	defer ethstatedb.TrieDB().Close()
```
joshklop, line 249: We should close the trie db with the `env.DeferErr` similar to the rawdb case above.

integrations/integrations.go
```diff
@@ -239,7 +240,7 @@ func startMonomerNode(
 		localdb.New(blockPebbleDB),
 		mempooldb,
 		txdb,
-		ethstatedb,
+		state.NewDatabase(ethstatedb),
```
joshklop, line 243: We should close the trie db similar to the case in `e2e/stack.go`.

node/node_test.go
```diff
@@ -34,9 +33,9 @@ func TestRun(t *testing.T) {
 	cometListener, err := net.Listen("tcp", cometHTTPAddress)
 	require.NoError(t, err)
 	app := testapp.NewTest(t, chainID.String())
-	ethstatedb := rawdb.NewMemoryDatabase()
+	ethstatedb := testutils.NewEthStateDB(t)
 	defer func() {
-		require.NoError(t, ethstatedb.Close())
+		require.NoError(t, ethstatedb.TrieDB().Close())
```
joshklop, line 38: Can we change `testutils.NewEthStateDB` to include

```go
t.Cleanup(func() {
	require.NoError(t, ethstatedb.TrieDB().Close())
})
```
that way we can remove this defer?

joshklop (changes requested):

Jesserc:
The latest commit addresses the review comments and also fixes #116

integrations/integrations.go
```diff
@@ -207,9 +209,12 @@ func startMonomerNode(
 		false,
 	)
 	if err != nil {
-		return fmt.Errorf("create eth state db: %v", err)
+		return fmt.Errorf("create raw db: %v", err)
 	}
-	env.DeferErr("close eth state db", ethstatedb.Close)
+	env.DeferErr("close rawDB db", rawDB.Close)
```
natebeauregard, line 214:
```suggestion
	env.DeferErr("close raw db", rawDB.Close)
```

node/node_test.go
natebeauregard: Could you add this change to `testutils.NewEthStateDB` like @joshklop  suggested so we don't need to include the cleanup steps in the test runner?

e2e/stack.go
```diff
@@ -243,8 +244,11 @@ func (s *Stack) runMonomer(ctx context.Context, env *environment.Env, genesisTim
 	env.DeferErr("close tx db", txdb.Close)
 	mempooldb := dbm.NewMemDB()
 	env.DeferErr("close mempool db", mempooldb.Close)
-	ethstatedb := rawdb.NewMemoryDatabase()
-	env.DeferErr("close eth state db", ethstatedb.Close)
+	rawDB := rawdb.NewMemoryDatabase()
+	env.DeferErr("close raw db", rawDB.Close)
+	trieDB := triedb.NewDatabase(rawDB, nil)
+	env.DeferErr("close trieDB", trieDB.Close)
+	ethstatedb := state.NewDatabaseWithNodeDB(rawDB, trieDB)
```
natebeauregard, line 251: nit: could you use lowercase `db` instead of `DB` to have parity with the rest of the database names here (and also in `integrations.go`)?

e2e/stack.go
```diff
@@ -243,8 +244,11 @@ func (s *Stack) runMonomer(ctx context.Context, env *environment.Env, genesisTim
 	env.DeferErr("close tx db", txdb.Close)
 	mempooldb := dbm.NewMemDB()
 	env.DeferErr("close mempool db", mempooldb.Close)
-	ethstatedb := rawdb.NewMemoryDatabase()
-	env.DeferErr("close eth state db", ethstatedb.Close)
+	rawDB := rawdb.NewMemoryDatabase()
+	env.DeferErr("close raw db", rawDB.Close)
```
natebeauregard, line 248: nit: It may be worth renaming `rawdb` to something like `rawstatedb` for readability, lmk what you think. This change would also apply in `integrations.go`

natebeauregard (commented):
Looking real good! Left a few small comments and suggestions

e2e/stack.go
joshklop: I think the raw db is more of a generic db backend, not specific to state. Probably better to keep it as `rawdb`.

Jesserc: Yeah, I think `rawdb` is fine too. But should it be in camelCase, `rawDB`, or `rawdb`?

natebeauregard: I don't have a strong opinion either way but we should use the same naming convention for all of the databases. It looks like we already have `txdb` and `mempooldb` as lowercase but it also looks like `blockPebbleDB` was recently added. Up to you on whether to rename

Jesserc: Yep, deleted my previous comment after noticing the convention. I'll rename it to `rawdb` and push the changes

Thanks @natebeauregard

node/node_test.go
natebeauregard: We'll still want to add this to the test utils. I think you should be able to refactor `testutils/utils.go#NewEthStateDB` like this to support the cleanup in the helper instead of the test itself.

```
func NewEthStateDB(t *testing.T) state.Database {
	rawstatedb := rawdb.NewMemoryDatabase()
	t.Cleanup(func() {
		require.NoError(t, rawstatedb.Close())
	})
	ethstatedb := state.NewDatabase(rawstatedb)
	t.Cleanup(func() {
		require.NoError(t, ethstatedb.TrieDB().Close())
	})
	return ethstatedb
}
```

e2e/stack.go
Jesserc: Unfortunately, this causes a lint error:

```bash
e2e/stack.go:247:2: importShadow: shadow of imported from 'github.com/ethereum/go-ethereum/core/rawdb' package 'rawdb' (gocritic)
        rawdb := rawdb.NewMemoryDatabase()
        ^
integrations/integrations.go:203:2: importShadow: shadow of imported from 'github.com/ethereum/go-ethereum/core/rawdb' package 'rawdb' (gocritic)
        rawdb, err := rawdb.NewPebbleDBDatabase(
        ^
make: *** [lint] Error 1
```

I have to rename it back to `rawDB`

natebeauregard (commented):
I left one more comment to move the trie db cleanup into the testutils helper, I'll approve/merge once it's addressed

e2e/stack.go
natebeauregard: Sounds good. You could alternatively use a different package reference for `rawdb` if you're getting a shadowing error but it's not that big of a deal

node/node_test.go
natebeauregard: Or even better - include both `t.Cleanup` statements in a single one :)

joshklop:
drive-by comment: (@natebeauregard feel free to override me if I'm missing something) Go linters typically encourage naming variables like `rawDB` instead of `rawdb` in my experience. The existing "codebase convention" to use `rawdb` is bad and shouldn't be followed. This will also help us prevent collisions with package names, which are lowercase.

natebeauregard:
> drive-by comment: (@natebeauregard feel free to override me if I'm missing something) Go linters typically encourage naming variables like `rawDB` instead of `rawdb` in my experience. The existing "codebase convention" to use `rawdb` is bad and shouldn't be followed. This will also help us prevent collisions with package names, which are lowercase.
> 
> Don't worry about it too much though. Just something to keep in mind in the future.

Yeah that makes sense. Should I make a small issue to rename `txdb`, `mempooldb`, etc. to avoid this confusion in the future?

joshklop:
> Should I make a small issue to rename txdb, mempooldb, etc. to avoid this confusion in the future?

If you feel strongly, feel free. imho if the linter doesn't complain, we don't need to waste time on it.

natebeauregard (approved):
LGTM! Lmk if you want me to squash them all into one commit for merge, otherwise feel free to organize these commits as you see fit and then tag me for merge

Jesserc:
> LGTM! Lmk if you want me to squash them all into one commit for merge, otherwise feel free to organize these commits as you see fit and then tag me for merge

Yes, you can squash them @natebeauregard

PR #115, opened 2024-08-08
Generate module configs with gocosmos and remove pulsar dependency

natebeauregard:
Now that Cosmos SDK module configs can be generated using gocosmos, all pulsar generated files are replaced with gocosmos generated files. To support module config generation with gocosmos, the `depinject` dependency needed to be updated to `v1.0.0` and the cosmos sdk core dependency needed to be updated to `v0.11.1`.

## Summary by CodeRabbit

- **New Features**
	- Streamlined build process by removing unnecessary dependencies and simplifying configurations.
  
- **Bug Fixes**
	- Improved build performance through the elimination of redundant commands related to protocol generation.

- **Chores**
	- Updated various dependencies to enhance compatibility, security, and performance within the module.


Makefile
```diff
@@ -33,8 +32,7 @@ lint:
 
 .PHONY: gen-proto
 gen-proto:
-	${GOBIN}/buf generate --template buf.gen.gocosmos.yaml --exclude-path $(PULSAR_PATHS)
-	${GOBIN}/buf generate --template buf.gen.pulsar.yaml --path $(PULSAR_PATHS)
+	${GOBIN}/buf generate --template buf.gen.gocosmos.yaml
```
joshklop, line 35: If memory serves me right, I think we can `buf.gen.gocosmos.yaml` to `buf.gen.yaml` and simplify this to `${GOBIN}/buf generate`. 

https://github.com/cosmos/cosmos-sdk/issues/20410#issuecomment-2115270916

joshklop (approved):
I love the diff. Thanks for doing this!

Makefile
natebeauregard: Good call, I'll update that! I'll probably leave this as three separate commits for merging, one for each dependency upgrade and one for removing the pulsar files, updating the Makefile, and generating the new module configs with gocosmos.

joshklop: > one for removing the pulsar files, updating the Makefile, and generating the new module configs with gocosmos.

Oh yeah, good idea to keep these changes together. Theoretically, the CI checks should pass on every individual commit merged to main.

natebeauregard: Just realized that I didn't update the Makefile to remove the `buf.gen.yaml` template, I'll include that in the next refactor PR

Issue #113, opened 2024-08-08
Sign Deposit transactions


NiloCK:
- Monomer receives deposit transaction(s) from the rollup node via the `forkchoiceUpdated` engine API.
- It packs these transactions into a cosmos friendly format.
- ... but the appchain's filtering mechanisms (AnteHandlers) reject this transaction for being unsigned.

A solution is for Monomer to hold some 'operator' key and sign the deposit transactions.

PR #66, opened 2024-07-24
Catch all potential errors in CheckTx

joshklop:
## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for transaction validation, providing clearer reporting on transaction outcomes.
	- Improved feedback for users regarding transaction conversion errors from Cosmos to Ethereum.

- **Bug Fixes**
	- Resolved issues with transaction validation that previously did not capture error details effectively.

PR #123, opened 2024-08-12
Sign L1 Deposit Txs

NiloCK:
Issue: https://github.com/polymerdao/monomer/issues/113

Context:

The changes from https://github.com/polymerdao/monomer/pull/66 enforce checks on the results of the appchain's `CheckTx` function. With those changes, the `CheckTx` workflow had been failing because of an absent signature field on the TXs.

This PR **reintroduces** the changes from #66, and adds:
- a `signer` struct to the engine
- the required fields for the ApplyL1Tx protobuf defintion

so that the adapters can optionally produce signed transactions.

## Summary by CodeRabbit

- **New Features**
	- Enhanced blockchain interaction with improved client creation and transaction signing capabilities.
	- Added support for application chain context management within the Node structure.
	- New transaction signing functionality introduced, enhancing transaction integrity and validation processes.

- **Bug Fixes**
	- Improved error handling and reporting for transaction validation.

- **Chores**
	- Simplified `ApplyL1TxsRequest` struct by removing the redundant `GetSigners` method.


engine/signer.go
```diff
@@ -0,0 +1,146 @@
...
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+	"github.com/cosmos/cosmos-sdk/crypto/keys/ed25519"
+	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
+	sdktx "github.com/cosmos/cosmos-sdk/types/tx"
+	"github.com/cosmos/cosmos-sdk/types/tx/signing"
+	authsigning "github.com/cosmos/cosmos-sdk/x/auth/signing"
+)
+
+type signer struct {
```
NiloCK, line 18: Maybe: `signer` could live in its own package as a child of `engine`. Would make this struct a bit safer.

engine/engine.go
```diff
@@ -31,6 +33,8 @@ type EngineAPI struct {
 	builder                  *builder.Builder
 	txValidator              TxValidator
 	blockStore               DB
+	appchainClient           *appchainClient.Context
```
NiloCK, line 36: Probably: should remove `appchainClient` as a member of the engine, since it gets pass on construction as a member of the signer.

proto/rollup/v1/msg.proto
```diff
@@ -15,9 +15,13 @@ service MsgService {
 
 // ApplyL1TxsRequest defines a message for all L1 system and user deposit txs
 message ApplyL1TxsRequest {
+  option (cosmos.msg.v1.signer) = "from_address";
```
natebeauregard, line 18: I believe that the `GetSigners` func in `gen/rollup/v1/helpers.go` can be removed now that we're using the `cosmos.msg.v1.signer` annotation

x/rollup/types/adapters.go
```diff
@@ -43,16 +45,26 @@ func AdaptPayloadTxsToCosmosTxs(ethTxs []hexutil.Bytes) (bfttypes.Txs, error) {
...
 			Messages: []*codectypes.Any{msgAny},
 		},
-	}).Marshal()
+	}
+
+	if signTx != nil {
+		err := signTx(&depositTx)
+		if err != nil {
+			return nil, fmt.Errorf("sign tx: %v", err)
+		}
```
natebeauregard, line 64: optional:
```suggestion
		if err := signTx(&depositTx); err != nil {
			return nil, fmt.Errorf("sign tx: %v", err)
		}
```

engine/signer.go
```diff
@@ -0,0 +1,146 @@
...
+	blankSig := signing.SignatureV2{
+		PubKey:   s.PubKey(),
+		Sequence: acc.GetSequence(),
+		Data: &signing.SingleSignatureData{
+			SignMode:  signing.SignMode_SIGN_MODE_DIRECT,
+			Signature: nil,
+		},
+	}
+
+	err = txBuilder.SetMsgs(tx.Body.Messages[0])
```
natebeauregard, line 95: nit: could you leave a comment here about why only the first message is passed through?

engine/signer.go
```diff
@@ -0,0 +1,146 @@
...
+							Mode: signing.SignMode_SIGN_MODE_DIRECT,
+						},
+					},
+				},
+				Sequence: acc.GetSequence(),
+			},
+		},
+		Fee: &sdktx.Fee{},
+	}
+	tx.Signatures = [][]byte{sig.Data.(*signing.SingleSignatureData).Signature}
```
natebeauregard, line 143: Is this logic to set the auth info and signatures already handled above by the [txBuilder.SetSignatures](https://github.com/polymerdao/monomer/pull/123/files#diff-7873bc6b8c386aec42ec5b42d310a7e2afacfc9f1a1ca12e97abe4df06b7eafcR116) call? If so, could these lines be removed?

engine/signer.go
```diff
@@ -0,0 +1,146 @@
...
+	pubKey        *cryptotypes.PubKey
+	address       *cryptotypes.Address
+	bech32Address *sdktypes.AccAddress
+}
+
+func NewSigner(appchainCtx *appchainClient.Context, privKey *ed25519.PrivKey) *signer {
+	return &signer{
+		appchainCtx: appchainCtx,
+		privKey:     privKey,
+	}
```
natebeauregard, line 30: curious - did you think about just setting the pubkey, address, and acc address during the `NewSigner` call instead of having them set as separate methods? Is there ever a case where we need some of those fields but not all of them?

Also, does `Address` need to be stored as a separate field or do we only care about `AccAddress` being stored on the `signer` struct? It looks like we can just pull the address when [signing](https://github.com/polymerdao/monomer/pull/123/files#diff-7873bc6b8c386aec42ec5b42d310a7e2afacfc9f1a1ca12e97abe4df06b7eafcR83) with `acc.GetAddress().String()`. Similarly, could the `PubKey` just be pulled off of the `privKey` with `privKey.PubKey()` whenever needed?

e2e/stack.go
```diff
@@ -266,3 +283,51 @@ func (s *Stack) runMonomer(ctx context.Context, env *environment.Env, genesisTim
...
+}
+
+func (mar mockAccount) GetAccountNumber() uint64 {
+	return 1
+}
+
+func (mar mockAccount) GetSequence() uint64 {
+	mar.seqNumber++
+	return mar.seqNumber
+}
```
natebeauregard, line 333: nit: could the receiver var be renamed from `mar` to something like `macc` for the `mockAccount` methods?

engine/signer.go
natebeauregard: Up to you! I think both approaches make sense

e2e/stack.go
```diff
@@ -266,3 +283,51 @@ func (s *Stack) runMonomer(ctx context.Context, env *environment.Env, genesisTim
 	}
 	return nil
 }
+
+type mockAccountRetriever struct {
```
natebeauregard, line 287: optional - the `mockAccountRetriever`/`mockAccount` structs and methods could be pulled out in a separate file for readability

engine/engine.go
```diff
@@ -44,13 +48,18 @@ func NewEngineAPI(
 	b *builder.Builder,
 	txValidator TxValidator,
 	blockStore DB,
+	appChainClient *appchainClient.Context,
 	metrics Metrics,
 ) *EngineAPI {
+	privKey := ed25519.GenPrivKeyFromSecret([]byte("monomer")) // TODO: configurable
```
natebeauregard, line 54: Will the `privKey` secret eventually be an environmental variable that's configured when monomer starts up? Or will it just be a random private key?

e2e/stack.go
```diff
@@ -266,3 +283,51 @@ func (s *Stack) runMonomer(ctx context.Context, env *environment.Env, genesisTim
...
+func (mar mockAccount) GetPubKey() cryptotypes.PubKey {
+	return nil
+}
+
+func (mar mockAccount) GetAccountNumber() uint64 {
+	return 1
+}
+
+func (mar mockAccount) GetSequence() uint64 {
+	mar.seqNumber++
```
natebeauregard, line 331: This is probably something that we don't need to worry about since it's just a mock account but I wanted to leave a comment just in case. If monomer errors out after a tx is signed but before it's sent will the sequence number get out of sync?

natebeauregard (commented):
Looking really good! Left a few small comments/questions

engine/signer.go
NiloCK: 4c3b6cc - sets all messages instead. Better to be general purpose.

There was a type-check wonkiness that I don't really understand which prevented using `tx.Body.Messages...`

e2e/stack.go
NiloCK: Shortened to `ma` here: a50ca48.

e2e/stack.go
NiloCK: Good idea. 9d1eafb

engine/engine.go
NiloCK: I think there could be use cases in either direction, so we'll probably want it consumable as an environment variable / command flag / whatever.

One might want to:
- sign deposit txs with a secret key (providing an authentication on the data)
- sign deposit txs with a known key (which makes them compatible with the current workings of appchain integrations, but provides no authentication)
- not sign deposit txs (not currently working, but could work in future with AnteHandler modifications or some modification on a custom tx type)

e2e/stack.go
NiloCK: Cosmos seqNumbers on transactions don't have to be in-order (`1, 2, 3, ...`), they just have to be increasing (`1, 2, 7, 8, ...`). I think this implementation will satisfy that.

engine/engine.go
NiloCK: e066e53

engine/signer.go
NiloCK: Went ahead and did this. 64b17f8, 6c9e5b7.

proto/rollup/v1/msg.proto
NiloCK: Great catch. 71ab0c6

engine/signer.go
NiloCK: Good points. `Address` is vestigial from the time I thought it was the actual field we needed. This stuff was cleaned up a little while `Signer` was moved into its own package.

engine/signer.go
NiloCK: There's a nuisance indirection happening here.

The `Sign` function here takes a tx and decorates it with the signature data.

The `txBuilder` instantiates its own internal tx and operates relative to that. `SetSignatures` operates on the data on that inernal tx - not the tx that we are decorating.

There might be some gains in clarity if `Sign` returned a tx.

NiloCK: Probably not worth it. The `tx` in a `txBuilder` is `github.com/cosmos/cosmos-sdk/x/auth/signing.Tx`, but the struct we're interested in manipulating / constructing is `github.com/cosmos/cosmos-sdk/types/tx.Tx`.

engine/signer/signer.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+}
+
+func (s *Signer) PubKey() cryptotypes.PubKey {
+	if s.pubKey == nil {
+		pubKey := s.privKey.PubKey()
+		s.pubKey = &pubKey
+	}
+	return *s.pubKey
+}
```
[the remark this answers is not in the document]
natebeauregard: It may be nice to avoid the indirection and use values instead of pointers for `PubKey` and `AccountAddress`, but I don't feel strongly either way.

engine/signer/signer.go
```diff
@@ -0,0 +1,143 @@
...
+					},
+				},
+				Sequence: seq,
+			},
+		},
+		Fee: &sdktx.Fee{},
+	}
+	tx.Signatures = [][]byte{sig.Data.(*signing.SingleSignatureData).Signature}
+
+	return nil
```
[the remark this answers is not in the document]
natebeauregard: Should `s.appchainCtx` be used instead of `context.Background` in the `SignWithPrivKey` func?

natebeauregard (approved):
LGTM! Left a couple follow-up questions based on the coderabbit comments but no need to re-review

engine/signer/signer.go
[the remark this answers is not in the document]
natebeauregard: Ahh nvm, I forgot that the `context` field passed in here isn't used down the line for the direct sign mode

PR #130, opened 2024-08-13
Refactor testapp module to adhere to Cosmos SDK suggested standards

### LINKED ISSUE #127: Refactor testapp module to follow suggested Cosmos SDK conventions
Refactor the testapp module to follow the [suggested Cosmos SDK conventions](https://docs.cosmos.network/main/build/building-modules/structure#structure). Listed below are some ideas for the refactor:

`testapp`

- Combine `get.proto` and `set.proto` into a single `tx.proto` file with a single service
- Generate `tx.pb.go` inside of `testapp/x/testmodule/types` instead of the `gen` directory
- Extract msg helper funcs from `gen/testapp/v1/helpers.go` into `testapp/x/testmodule/types/codec.go` (`RegisterInterfaces`) and `testapp/x/testmodule/types/msgs.go` (`ValidateBasic`, `GetSigners`, `Type`, `Route`, etc)

natebeauregard:
Closes #127 

Refactors the testapp module to follow the [suggested Cosmos SDK conventions](https://docs.cosmos.network/main/build/building-modules/structure#structure).

- Combines `get.proto` and `set.proto` into `query.proto` and `tx.proto` files
- Generates pbs inside of `testapp/x/testmodule/types` instead of the `gen` directory
- Extracts msg helper funcs from `gen/testapp/v1/helpers.go` into `testapp/x/testmodule/types/codec.go` (`RegisterInterfaces`) and `testapp/x/testmodule/types/msgs.go` (`ValidateBasic`)

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `QueryService` for improved data retrieval operations.
	- Added a `MsgService` for handling message transactions.
	- Added message structure `SetRequest` with a placeholder validation method.

- **Improvements**
	- Enhanced organization of protocol buffer generation with a new directory for generated types.
	- Updated import paths for better modularity within the application.
	- Refactored request and response types for improved consistency across the application.

- **Bug Fixes**
	- Resolved inconsistencies in request and response types for enhanced functionality across the application.


testapp/helpers.go
```diff
@@ -128,7 +128,7 @@ func (a *App) StateDoesNotContain(t *testing.T, height uint64, kvs map[string]st
 		}).Marshal()
 		require.NoError(t, err)
 		resp, err := a.Query(context.Background(), &abcitypes.RequestQuery{
-			Path:   QueryPath, // TODO is there a way to find this programmatically?
+			Path:   QueryPath,
```
natebeauregard, line 131: I looked into generating and passing through the query path programmatically, however since this use case is only for this helper file I decided it wasn't worth the overhead to refactor.

For visibility, there were two potential approaches that I looked into:
1. Use the `ServiceName` and "Get" `MethodName` from [`_QueryService_serviceDesc`](https://github.com/polymerdao/monomer/pull/130/files#diff-77a41b8c9fc4875149ea5fc72fb2487b4d02031b00df1d3de313d3934cb1418bR211-R222) to dynamically build this path. However, the method name field is numerically indexed so our path would break if "Get" isn't the first method name.
2. [Create](https://github.com/polymerdao/monomer/pull/130/files#diff-77a41b8c9fc4875149ea5fc72fb2487b4d02031b00df1d3de313d3934cb1418bR163) a new query service client and then [call](https://github.com/polymerdao/monomer/pull/130/files#diff-77a41b8c9fc4875149ea5fc72fb2487b4d02031b00df1d3de313d3934cb1418bR167) the `Get` method to return the get response. However, this would require a good bit of additional boilerplate code to create the `grpc1.ClientConn` used to build the query service client.

joshklop: Hmm interesting. Thanks for investigating! This will be valuable information when we add more APIs to the Cosmos SDK app in the `integrations` package.

comet/comet_test.go
```diff
@@ -17,9 +17,9 @@ import (
 	"github.com/polymerdao/monomer"
 	"github.com/polymerdao/monomer/app/peptide/txstore"
 	"github.com/polymerdao/monomer/comet"
-	"github.com/polymerdao/monomer/gen/testapp/v1"
 	"github.com/polymerdao/monomer/mempool"
 	"github.com/polymerdao/monomer/testapp"
+	testapptypes "github.com/polymerdao/monomer/testapp/x/testmodule/types"
```
joshklop, line 22: nit: should this be `testmoduletypes`?

scripts/gen-proto.sh
```diff
@@ -14,4 +15,6 @@ buf generate
 cp -r $GEN_DIR/rollup/v1/* $ROLLUP_DIR/types
 rm -rf $GEN_DIR/rollup/v1
 
-# TODO: move the testapp module message types to the testapp/x/testmodule module
+# move the generated testapp module message types to the testapp/x/testmodule module
+cp -r $GEN_DIR/testapp/v1/* $TESTMODULE_DIR/types
+rm -rf $GEN_DIR/testapp/v1
```
joshklop, line 20: Can we remove the entire `$GEN_DIR` now?

joshklop (approved):
Thanks for doing this. Do we need to update any of the `buf lint` settings to match the sdk conventions?

scripts/gen-proto.sh
natebeauregard: The ex-pulsar `module.pb.go` files are still generated there, however I felt like they didn't really make sense to include in the module itself. The Cosmos SDK repo includes their generated `module.pulsar.go` files in a separate `api` directory, would renaming the `gen` directory to `api` make sense here as well?

natebeauregard:
> Thanks for doing this. Do we need to update any of the `buf lint` settings to match the sdk conventions?

@joshklop I previously ran into lint issues while trying to rename the `x/rollup` message names to be prefixed with `Msg...` to follow the cosmos sdk [naming convention](https://docs.cosmos.network/main/build/building-modules/messages-and-queries#msg-services), however buf didn't like that. The naming conventions in Cosmos aren't really enforced however, they're more meant for guidance. I don't feel too strongly about this one way or another though, do you think it's worth looking into updating buf to allow the `Msg` prefix?

scripts/gen-proto.sh
joshklop: Thanks, I forgot about those... let's leave it for now.

joshklop:
@natebeauregard Have we tried porting these configs? https://github.com/cosmos/cosmos-sdk/blob/0fe3115dcf2e42dac29364a70cca71d52df2a443/proto/buf.yaml#L13-L26

natebeauregard:
> @natebeauregard Have we tried porting these configs? https://github.com/cosmos/cosmos-sdk/blob/0fe3115dcf2e42dac29364a70cca71d52df2a443/proto/buf.yaml#L13-L26

I haven't looked into fixing the cosmos sdk naming lint issues yet if we want the `Msg...` prefix but that looks pretty promising! What's your opinion on updating the message names? I feel like it would be nice to have to match existing sdk modules, if you're not opposed I can make these changes in this PR real quick

joshklop:
@natebeauregard Yeah that would be great if you have the time. Might as well take care of it now to establish the convention.

natebeauregard:
@joshklop On second thought it might make sense to include in a separately scoped PR if we're going to change the `x/rollup` messages as well as the testapp messages. I'll make a new github issue for this and mark it as a good first issue if the interns want to take a stab at it while I'm gone

joshklop:
> @joshklop On second thought it might make sense to include in a separately scoped PR if we're going to change the `x/rollup` messages as well as the testapp messages. I'll make a new github issue for this and mark it as a good first issue if the interns want to take a stab at it while I'm gone

Good idea

PR #131, opened 2024-08-13
Clean up Monomer EVM chain config

### LINKED ISSUE #104: Investigate SSTORE access list EVM execution bug with BerlinBlock/LondonBlock
Currently, the following error is occuring in the `builder_test` whenever the `BerlinBlock` and `LondonBlock` are enabled in the Monomer EVM: `evm: address was not present in access list during sstore op`. Those upgrades are [disabled](https://github.com/polymerdao/monomer/blob/89a182fcad9f7e0ae94de3b3defb93ea409f8a96/evm/evm.go#L34-L36) for now in the Monomer EVM. Firstly, we should investigate why the error is occurring and whether we need the `LondonBlock` and `BerlinBlock` to be enabled for our Monomer EVM execution to have parity with existing Optimism EVM execution. Alternatively, we'll need to add in an access list for the Monomer EVM.

The Monomer EVM doesn't account for gas and only calls the `L2ToL1MessagePasser` and `L2ApplicationStateRootProvider` contracts, so as long as all changes in the London and Berlin upgrades are in reference to gas changes or opcodes that we don't use in the Monomer EVM we may just want to omit those upgrades from Monomer.

Additionally, we should introduce a shared variable for the chain config used by both the [Monomer EVM](https://github.com/polymerdao/monomer/blob/89a182fcad9f7e0ae94de3b3defb93ea409f8a96/evm/evm.go#L26) and [RPCMarshalBlock](https://github.com/polymerdao/monomer/blob/89a182fcad9f7e0ae94de3b3defb93ea409f8a96/eth/internal/ethapi/helpers.go#L20).

Jesserc:
Closes #104 
## Context
Berlin and London hardforks primarily introduce gas-related changes. Since Monomer doesn't currently account for gas, these forks may be unnecessary for the Monomer EVM.

## Changes
- Remove commented Berlin and London fork activations in [Monomer EVM](https://github.com/polymerdao/monomer/blob/89a182fcad9f7e0ae94de3b3defb93ea409f8a96/evm/evm.go#L34-L36)
- There is also no need for shared EVM config variable between [RPCMarshalBlock](https://github.com/polymerdao/monomer/blob/89a182fcad9f7e0ae94de3b3defb93ea409f8a96/eth/internal/ethapi/helpers.go#L20) and Monomer EVM, since `RPCMarshalBlock` needs both Berlin and London fork activations but Monomer EVM doesn't

## Summary by CodeRabbit

- **Chores**
    - Cleaned up commented-out code related to outdated block references, streamlining the codebase for improved clarity.

### DIFF
--- a/evm/evm.go
+++ b/evm/evm.go
@@ -31,12 +31,9 @@ func NewEVM(ethState vm.StateDB, header *monomer.Header) (*vm.EVM, error) {
 		PetersburgBlock:     new(big.Int),
 		IstanbulBlock:       new(big.Int),
 		MuirGlacierBlock:    new(big.Int),
-		// TODO: investigate SSTORE access list EVM execution bug with BerlinBlock/LondonBlock
-		// BerlinBlock:        new(big.Int),
-		// LondonBlock:        new(big.Int),
-		ArrowGlacierBlock:  new(big.Int),
-		GrayGlacierBlock:   new(big.Int),
-		MergeNetsplitBlock: new(big.Int),
+		ArrowGlacierBlock:   new(big.Int),
+		GrayGlacierBlock:    new(big.Int),
+		MergeNetsplitBlock:  new(big.Int),
 
 		BedrockBlock: new(big.Int),
 		RegolithTime: utils.Ptr(uint64(0)),


Jesserc:
cc @natebeauregard

natebeauregard (approved):
LGTM!

PR #135, opened 2024-08-14
CI: enforce protobuf pipeline

NiloCK:
This repo contains generated code as checked-in source.

This change adds a CI check that the pipeline for generated (`source`, `buildsteps`, `outputs`) code is intact, and prevents accidentally merging, eg, changes to a protobuf definition file without corresponding changes to generated outputs.

Should prevent issues as described in #133

## Summary by CodeRabbit

- **New Features**
	- Introduced a new job to verify generated code, ensuring it is up-to-date and prompting users for local actions if discrepancies arise.
	- Added a new testing job to enhance the overall CI/CD process.
  
- **Improvements**
	- Renamed existing jobs for clarity and better functionality, improving the development workflow efficiency.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,7 +46,7 @@ jobs:
         with:
           config: .testcoverage.yml
 
-  unit:
+  short:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,15 +7,6 @@ on:
   workflow_dispatch:
 
 jobs:
-  short:
-    runs-on: macos-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
-        with:
-          go-version-file: go.mod
-      - run: make test
-
   devrequirements:
     runs-on: macos-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -7,23 +7,32 @@ on:
...
-  short:
+  devrequirements:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
-      - run: make test
+      - run: make install-buf
```
natebeauregard, line 17: Is the `devrequirements` job necessary any more now that the `verify-generated-code` job also runs `make install-buf`?

.github/workflows/test.yml
```diff
@@ -37,6 +46,15 @@ jobs:
         with:
           config: .testcoverage.yml
 
+  unit:
```
natebeauregard, line 49: optional - I believe `make test` includes some integration tests and only excludes the e2e tests. It may be worth either renaming this from "unit" to something more generic or making a new testing job for integration tests.

natebeauregard (approved):
LGTM

.github/workflows/test.yml
NiloCK: This is a good observation. Will switch back.

.github/workflows/test.yml
NiloCK: Maybe not, but I sort of like the explicitness of it. Will fence sit on it for a bit.

PR #146, opened 2024-08-15
Refactor error handling and update tests

AnkushinDaniil:
Refactored the error handling in the `AdaptCosmosTxsToEthTxs` function to return a specific error message when unmarshaling binary transactions fails. Also updated the `builder_test` and `localdb_test` files to reflect the changes.


AnkushinDaniil:
After merging this PR I'll update guard against depTxs

Issue #144, opened 2024-08-15
Refactor e2e tests


NiloCK:
The e2e test has become somewhat unwieldy - lots of setup (stack, clients), triggering operations, and assertions against expected outcomes intermixed. It's easy to lose track of the link between assertions and inputs.

There are also some [timing hacks](https://github.com/polymerdao/monomer/blob/7705f8abfb8fb595d6f2d64f74e6017e83c9d50f/e2e/stack_test.go#L176-L184) present. 

Proposed refactor:
- modify the `stack` constructor / runner to return a richer set of interfacing components (eg, `appchainClient`, `l1RPCClient`)
- Create a structured format, against the above interfaces, with
  - a test name
  - a `triggering event` executing function
  - an `expected outcome` assertion function (the [check for eth minting](https://github.com/polymerdao/monomer/blob/85c2a3c97e9ebc7526d14838c0d6b7aa46869dea/e2e/stack_test.go#L237) in #140 is roughly such an assertion)
- Migrate existing functionality to the structured format

to do:
- ~[ ] remove hardcoded ports from various services~ (nah - more headache than upside)

joshklop:
I really like the first bullet point. We could have the `Stack` do almost all the setup (no need to accept a ton of URLs in the constructor) as long as it returns the proper clients. This is similar to the `System` struct in the `op-e2e` tests I believe.

I'm a little more wary of the second point. This sounds like a lot of callbacks. Perhaps I'm misunderstanding though.

Can we have a PR that implements the first suggestion while we discuss the remaining items?

joshklop:
@NiloCK Instead of having a separate constructor and `Run` method for the `Stack`, maybe we can have a single `Setup(ctx, env, config) (*Stack, error)` where the `Stack` struct contains the clients. Is this what you were thinking?

NiloCK:
> I'm a little more wary of the second point. This sounds like a lot of callbacks. Perhaps I'm misunderstanding though.

I've probably been unclear or misunderstood the implications of my own suggestion, but I despise callbacks and don't usually intend to design toward them.

In this case, think _imperative functions that operate on a stack_, not _stacks that consume imperative functions_.



> @NiloCK Instead of having a separate constructor and `Run` method for the `Stack`, maybe we can have a single `Setup(ctx, env, config) (*Stack, error)` where the `Stack` struct contains the clients. Is this what you were thinking?

Yup.

PR #149, opened 2024-08-15
[e2e]: Construct & return clients with `stack.Setup()`

NiloCK:
PR is an incremental part of a refactor plan for the e2e tests.

See #144.

This PR addresses the first bullet point.

## Summary by CodeRabbit

- **New Features**
	- Enhanced configuration management for end-to-end testing, allowing for better integration and interaction with Layer 1 and Layer 2 clients.
	- Introduction of a new URL helper function for improved URL construction and error handling.

- **Bug Fixes**
	- Improved error handling during setup to ensure valid configuration paths.

- **Refactor**
	- Streamlined test configurations by reducing hardcoded values and centralizing setup logic for better readability and maintainability.

### DIFF
--- a/e2e/stack.go
+++ b/e2e/stack.go
@@ -41,11 +41,12 @@ type EventListener interface {
 }
 
 type StackConfig struct {
-	L1URL    *url.URL
-	RUConfig *rollup.Config
-	Operator L1User
-	Users    []L1User
-	L1Client *L1Client
+	L1URL         *url.URL
+	RUConfig      *rollup.Config
+	Operator      L1User
+	Users         []L1User
+	L1Client      *L1Client
+	MonomerClient *MonomerClient
 }
 
 type stack struct {
@@ -219,11 +220,12 @@ func (s *stack) run(ctx context.Context, env *environment.Env) (*StackConfig, er
 	}
 
 	return &StackConfig{
-		L1URL:    l1url,
-		L1Client: l1,
-		RUConfig: rollupConfig,
-		Operator: l1users[0],
-		Users:    l1users[1:],
+		L1URL:         l1url,
+		L1Client:      l1,
+		MonomerClient: monomerClient,
+		RUConfig:      rollupConfig,
+		Operator:      l1users[0],
+		Users:         l1users[1:],
 	}, nil
 }
 
--- a/e2e/stack.go
+++ b/e2e/stack.go
@@ -67,7 +67,11 @@ type L1User struct {
 	PrivateKey *ecdsa.PrivateKey
 }
 
-// Setup assumes all ports are available and that all paths exist and are valid.
+// Setup creates and runs a new stack for end-to-end testing.
+//
+// It assumes availability of hard-coded local URLs for the Monomer engine, Comet, and OP node.
+//
+// It returns a StackConfig with L1 and L2 clients, the rollup config, and operator and user accounts.
 func Setup(
 	ctx context.Context,
 	env *environment.Env,
--- a/e2e/stack.go
+++ b/e2e/stack.go
@@ -41,8 +41,8 @@ type EventListener interface {
 }
 
 type StackConfig struct {
-	L1URL *url.URL
-	*rollup.Config
+	L1URL    *url.URL
+	RUConfig *rollup.Config
 	Operator L1User
 	Users    []L1User
 }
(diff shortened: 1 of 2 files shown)


e2e/stack.go
```diff
@@ -41,16 +43,19 @@ type EventListener interface {
...
+	L1Client      *L1Client
+	L2Client      *bftclient.HTTP
+	MonomerClient *MonomerClient
 }
 
-type Stack struct {
-	monomerEngineURL *url.URL
-	monomerCometURL  *url.URL
-	opNodeURL        *url.URL
+type stack struct {
```
joshklop, line 55: Do you think it would be better to remove this struct? It doesn't serve any purpose.

e2e/stack.go
```diff
@@ -41,16 +43,19 @@ type EventListener interface {
...
-	*rollup.Config
-	Operator L1User
-	Users    []L1User
+	L1URL         *e2eurl.URL
+	RUConfig      *rollup.Config
+	Operator      L1User
+	Users         []L1User
+	L1Client      *L1Client
+	L2Client      *bftclient.HTTP
+	MonomerClient *MonomerClient
```
joshklop, line 52: Can remove the `L1URL` and `RUConfig`?

e2e/stack.go
```diff
@@ -62,18 +67,40 @@ type L1User struct {
...
+		return nil, fmt.Errorf("new l1 url: %v", err)
+	}
+	monomerCometURL, err := newURL("http://127.0.0.1:8890")
+	if err != nil {
+		return nil, fmt.Errorf("new l1 url: %v", err)
+	}
+	opNodeURL, err := newURL("http://127.0.0.1:8891")
+	if err != nil {
+		return nil, fmt.Errorf("new l1 url: %v", err)
+	}
```
joshklop, line 101: All of the error messages are the same.
Can the port in each address be `0`?

e2e/stack.go
```diff
@@ -285,3 +323,15 @@ func (s *Stack) runMonomer(ctx context.Context, env *environment.Env, genesisTim
...
+	stdURL, err := url.Parse(address)
+	if err != nil {
+		return nil, fmt.Errorf("parse URL: %v", err)
+	}
+	resultURL, err := e2eurl.Parse(stdURL)
+	if err != nil {
+		return nil, fmt.Errorf("parse e2e URL: %v", err)
+	}
+	return resultURL, nil
+}
```
joshklop, line 337: `e2eurl.ParseString`?

e2e/stack.go
NiloCK: It would probably be better to remove, but it currently still shepherds some data into the constructor for `e2e.OPStack`, and make for a bigger change set than necessary / convenient for the immediate purpose of cleaning up the e2e tests.

e2e/stack.go
NiloCK: Error messages fixed here: 35bf818

The `0` port idea being that each service would look up and assume some available port at runtime?

e2e/stack.go
NiloCK: `L1URL` is removed here: 22dd6db

`RUConfig` has data still used to construct the op `portal` contract bindings. I do intend to swallow that up into `stack.go`, as the client constructions were. After that, the `RUConfig` can be removed as well.

Include in this PR? Or follow-up?

joshklop: follow-up sounds good, thanks :+1:

e2e/stack.go
joshklop: > The 0 port idea being that each service would look up and assume some available port at runtime?

kernel automatically assigns unused port when `0` is provided

e2e/stack.go
joshklop: sgtm, better for follow-up then

joshklop (approved):

joshklop:
preemptively approved, pending resolution on the port discussion.

e2e/stack.go
NiloCK: I had tried using `:0` initially, and the test hangs.

The reason is that (some of?) the values `monomerEngineURL`, `monomerCometURL`, and `opNodeURL` are read more than once - first to initialize a listener, then later to connect a client to that address.

If we'd like to dynamically assign the cleanest thing is probably to remove these values altogether, and force correct piping of address data down the line.

joshklop: Ahhh that's what I was forgetting. Disregard me then. This is why it's nice when APIs accept `Listener`s so we can learn the address that the kernel assigns.

Issue #134, opened 2024-08-13
Add unit tests for the x/rollup module


natebeauregard:
Add unit test coverage to assert behavior for the `x/rollup` messages and the related helper functions involved in the deposit and withdrawal flow.

### Unit

- `deposits`
    - `processL1SystemDepositTx`
    - `processL1UserDepositTxs`
    - `mintETH`
- `withdrawals`
    - `burnETH`

### Integration

- `msg_server` (may want these to be unit instead - will depend on if we want to mock the bank/mint keepers)
    - `ApplyL1Txs`
    - `InitiateWithdrawal`

Issue #160, opened 2024-08-20
Investigate whether Monomer should store L1 attributes


joshklop:
Questions to answer:

* What purpose do they serve on traditional OP Stack chains?
* Does that purpose make sense for Monomer-enabled applications?
* If we need to make changes to the `x/rollup` module, how should we go about it?

PR #161, opened 2024-08-21
Add x/rollup unit tests

natebeauregard:
Closes: https://github.com/polymerdao/monomer/issues/134

This PR is heavily influenced by the [Cosmos SDK testing docs](https://docs.cosmos.network/main/build/building-modules/testing) and existing Cosmos SDK/Osmosis unit tests.

- Use expected keeper interface for `x/rollup` bankkeeper
- Generate expected keeper mocks with mockgen
  - Adds new makefile commands to install mockgen and to generate expected keeper mocks for use in `x/rollup` unit tests.
- Adds unit test coverage to assert behavior for the `x/rollup` messages and the related helper functions involved in the deposit and withdrawal flow.

## Summary by CodeRabbit

- **New Features**
	- Introduced `install-mockgen` and `gen-mocks` targets in the build process to facilitate mock generation for testing.
	- Added a `BankKeeper` interface to define methods for managing coin transfers, minting, and burning within the rollup module.
	- Implemented comprehensive unit tests for the `Keeper` component and its methods, enhancing the reliability of transaction processing and withdrawals.
	- Refactored the `GenerateEthTxs` function to encapsulate block generation logic, improving modularity and readability.

- **Bug Fixes**
	- Enhanced error handling in transaction processing functions to ensure proper responses to invalid inputs.

- **Documentation**
	- Updated comments and naming conventions for functions to improve clarity and consistency across the codebase.


x/rollup/keeper/deposits.go
```diff
@@ -97,15 +98,17 @@ func (k *Keeper) processL1UserDepositTxs(ctx sdk.Context, txs [][]byte) error {
...
+func (k *Keeper) MintETH(ctx sdk.Context, addr sdk.AccAddress, amount sdkmath.Int) error { //nolint:gocritic // hugeParam
 	coin := sdk.NewCoin(types.ETH, amount)
 	if err := k.bankkeeper.MintCoins(ctx, types.ModuleName, sdk.NewCoins(coin)); err != nil {
 		return fmt.Errorf("failed to mint deposit coins to the rollup module: %v", err)
 	}
 	if err := k.bankkeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, addr, sdk.NewCoins(coin)); err != nil {
 		return fmt.Errorf("failed to send deposit coins from rollup module to user account %v: %v", addr, err)
 	}
+
+	// TODO: move event emission to msg_server.go
```
natebeauregard, line 111: Created issue https://github.com/polymerdao/monomer/issues/162

x/rollup/keeper/deposits.go
```diff
@@ -24,26 +24,27 @@ func (k *Keeper) setL1BlockInfo(ctx sdk.Context, info derive.L1BlockInfo) error
...
 		return types.WrapError(err, "marshal L1 block info")
 	}
-	if err := k.storeService.OpenKVStore(ctx).Set([]byte(types.KeyL1BlockInfo), infoBytes); err != nil {
+	if err = k.storeService.OpenKVStore(ctx).Set([]byte(types.KeyL1BlockInfo), infoBytes); err != nil {
 		return types.WrapError(err, "set")
 	}
 	return nil
 }
 
+// TODO: include the logic to also store the L1 block info by blockhash in setL1BlockInfo and remove setL1BlockHistory
```
natebeauregard, line 33: Created issue https://github.com/polymerdao/monomer/issues/163

joshklop:
Instead of refactoring to use an interface with our own mock, what do you think about these options:

- Bank module mock from the SDK https://github.com/cosmos/cosmos-sdk/blob/v0.50.6/x/bank/testutil/expected_keepers_mocks.go
- Integration test utils https://github.com/cosmos/cosmos-sdk/blob/v0.50.6/testutil/integration/example_test.go. [Here](https://github.com/cosmos/cosmos-sdk/blob/v0.50.6/tests/integration/bank/keeper/deterministic_test.go) is an example of how it is used for the bank module.

I tend to prefer just testing the actual behavior rather than mocking (I guess the SDK calls that an "integration test" and not a unit test, oh well). But I can be persuaded otherwise if the integration test utils are too cumbersome.

natebeauregard:
> Instead of using our own mock, what do you think about these options:
> 
> * Bank module mock from the SDK https://github.com/cosmos/cosmos-sdk/blob/v0.50.6/x/bank/testutil/expected_keepers_mocks.go
> * Integration test utils https://github.com/cosmos/cosmos-sdk/blob/v0.50.6/testutil/integration/example_test.go. [Here](https://github.com/cosmos/cosmos-sdk/blob/v0.50.6/tests/integration/bank/keeper/deterministic_test.go) is an example of how it is used for the bank module.
> 
> I tend to prefer just testing the actual behavior rather than mocking (I guess the SDK calls that an "integration test" and not a unit test, oh well). But I can be persuaded otherwise if the integration test utils are too cumbersome.
> 
> My thinking is that we might as well just go straight to the integration test if it isn't too much effort, rather than having an extra unit test that involves a lot of mocking.

@joshklop 

**Bank module mock from the SDK**
I'm a little confused by this option, the linked `x/bank/testutil/expected_keepers_mocks.go` is not meant to be used externally for mocking the bank keeper. It's instead used by the bank keeper unit tests to mock the account keeper so it's easy to get full test coverage without needing to implement every single account keeper error/edge case in integration tests.

**Integration test utils**
Yeah it would be nice to have integration tests as well to assert the rollup and bank modules interact together correctly, but I don't think that all of our testing for the x/rollup module should be in an integration test. Unit tests are nice since they're only scoped to the rollup module and are easy to get full test coverage for through mocks. As you mentioned in your comment however, the main drawback of unit tests is not being able to look at the full picture of any related modules interacting with each other in a test app.

After re-reviewing the added unit tests in this PR and re-reading through the [Testing docs](https://docs.cosmos.network/main/build/building-modules/testing) and [ADR 059](https://docs.cosmos.network/main/build/architecture/adr-059-test-scopes.html), I don't think that any of them would qualify to split out into integration tests. I'll start working on a separate PR to add a rollup module integration test to the `/tests/integration/` directory that has the full deposit -> withdrawal flow with a real bank keeper and a few scattered error cases. Please lmk if you want to sync up to discuss this in case there was any misunderstanding.

joshklop:
@natebeauregard Sounds good, appreciate the descriptions. Let's go with what you suggested.

x/rollup/types/expected_keepers.go
```diff
@@ -0,0 +1,16 @@
+package types
+
+import (
+	"context"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// BankKeeper defines the expected bank keeper interface used in the x/rollup module
+type BankKeeper interface {
```
joshklop, line 10: Are there reasons to put the interface in the `types` package instead of in the `keeper` package? Is this an SDK convention?

testutils/utils.go
```diff
@@ -65,17 +65,11 @@ func NewLocalMemDB(t *testing.T) *localdb.DB {
...
-		Difficulty: common.Big0,
-		Number:     big.NewInt(0),
-		Time:       timestamp,
-	}, nil, nil, nil, trie.NewStackTrie(nil))
+	l1Block := GenerateL1Block()
 	l1InfoRawTx, err := derive.L1InfoDeposit(&rollup.Config{
 		Genesis:   rollup.Genesis{L2: eth.BlockID{Number: 0}},
 		L2ChainID: big.NewInt(1234),
-	}, eth.SystemConfig{}, 0, eth.BlockToInfo(l1Block), timestamp)
+	}, eth.SystemConfig{}, 0, eth.BlockToInfo(l1Block), uint64(0))
```
joshklop, line 72: Instead of using `uint64(0)`, can we use something like `l1Block.Timestamp()`? That way if we change the timestamp in the future we can avoid weird bugs.

x/rollup/keeper/deposits.go
```diff
@@ -61,8 +62,8 @@ func (k *Keeper) processL1SystemDepositTx(ctx sdk.Context, txBytes []byte) (*der
 	return l1blockInfo, nil
 }
 
-// processL1UserDepositTxs processes the L1 user deposit txs and mints ETH to the user's cosmos address.
-func (k *Keeper) processL1UserDepositTxs(ctx sdk.Context, txs [][]byte) error { //nolint:gocritic // hugeParam
+// ProcessL1UserDepositTxs processes the L1 user deposit txs and mints ETH to the user's cosmos address.
+func (k *Keeper) ProcessL1UserDepositTxs(ctx sdk.Context, txs [][]byte) error { //nolint:gocritic // hugeParam
```
joshklop, line 66: This expands the API. Is there a reason to do this other than testing? Generally, it's better to test all behavior via the public interface (I think one of the only times we eschew that convention in our codebase is in the `comet` package for the paginate functionality).

x/rollup/types/expected_keepers.go
natebeauregard: It's just a Cosmos SDK quirk. It's mentioned as being in the types package in the [Folder Structure docs](https://docs.cosmos.network/main/build/building-modules/structure#structure).

The [keeper docs](https://docs.cosmos.network/main/build/building-modules/keeper#type-definition) touch on it a little more:
> An expected keeper is a keeper external to a module that is required by the internal keeper of said module. External keepers are listed in the internal keeper's type definition as interfaces. These interfaces are themselves defined in an expected_keepers.go file in the root of the module's folder. In this context, interfaces are used to reduce the number of dependencies, as well as to facilitate the maintenance of the module itself.

x/rollup/keeper/deposits.go
natebeauregard: Nope, they were only exposed for test access. There were a bunch of modules in the Cosmos SDK that had similarly exposed funcs for testing purposes only, but I agree that it's not the best precedent to set.

Do you think it would be better to keep the separate `deposits_test.go` and `withdrawals_test.go` tests but move them into the `keeper` package to access the unexported funcs or delete them and try to test every error/edge case through `msg_server_test.go`?

x/rollup/types/expected_keepers.go
joshklop: Great reference, thanks. Not a fan of the convention myself, but probably best to follow it here.

x/rollup/keeper/msg_server_test.go
```diff
@@ -0,0 +1,157 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/ethereum-optimism/optimism/op-node/rollup/derive"
+	"github.com/ethereum-optimism/optimism/op-service/eth"
+	"github.com/golang/mock/gomock"
+	"github.com/polymerdao/monomer/testutils"
+	"github.com/polymerdao/monomer/x/rollup/keeper"
+	"github.com/polymerdao/monomer/x/rollup/types"
+)
+
+func (s *KeeperTestSuite) TestApplyL1Txs() {
```
joshklop, line 16: Is the SDK convention to use a suite? We don't use suites elsewhere in the codebase. Any thoughts on the advantages and disadvantages?

x/rollup/keeper/deposits.go
joshklop: Any thoughts on testing all edge cases via the public interface `ApplyL1Txs`? I haven't investigated the tests in detail, so forgive me if I'm missing something obvious.

natebeauregard: Noting for visibility - we talked during our daily call and addressed this question, we'll move all unit tests into `msg_server_test.go` since we don't anticipate many more test cases being added

x/rollup/keeper/msg_server_test.go
natebeauregard: Noting for visibility - we talked during our daily call and addressed this question. Test suites are nice for cosmos sdk modules since you can store all relevant keeper/store info on them and reset the stores/keepers before every subtest automatically

x/rollup/keeper/deposits.go
```diff
@@ -108,7 +108,7 @@ func (k *Keeper) MintETH(ctx sdk.Context, addr sdk.AccAddress, amount sdkmath.In
 		return fmt.Errorf("failed to send deposit coins from rollup module to user account %v: %v", addr, err)
 	}
 
-	// TODO: move event emission to msg_server.go
+	// TODO: only emit sdk.EventTypeMessage once per message
```
natebeauregard, line 111: Noticed a bug with the message emission for `ApplyL1Txs`, I updated https://github.com/polymerdao/monomer/issues/162 to reflect this

x/rollup/keeper/msg_server_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	for name, test := range tests {
+		s.Run(name, func() {
+			if test.setupMocks != nil {
+				test.setupMocks()
+			}
+			s.mockMintETH()
+
+			var resp *types.MsgApplyL1TxsResponse
+			resp, err = keeper.NewMsgServerImpl(s.rollupKeeper).ApplyL1Txs(s.ctx, test.msg)
```
joshklop, line 126: Why the extra `var` and not `:=`?

x/rollup/keeper/msg_server_test.go
```diff
@@ -0,0 +1,236 @@
...
+			},
+			shouldError: true,
+		},
+		"contract creation tx passed in as user deposit tx": {
+			msg: &types.MsgApplyL1Txs{
+				TxBytes: [][]byte{l1AttributesTxBz, contractCreationTxBz},
+			},
+			shouldError: true,
+		},
+		"one successful l1 user deposit tx and an invalid tx passed in as user deposit txs": {
```
joshklop, line 92: nit: not sure if this is correct, but seems right
```suggestion
		"one valid l1 user deposit tx and an invalid tx passed in as user deposit txs": {
```

x/rollup/keeper/msg_server_test.go
```diff
@@ -0,0 +1,236 @@
...
+	depositTxBz, err := depositTx.MarshalBinary()
+	s.Require().NoError(err)
+	cosmosEthTxBz, err := cosmosEthTx.MarshalBinary()
+	s.Require().NoError(err)
+	contractCreationTxBz, err := contractCreationTx.MarshalBinary()
+	s.Require().NoError(err)
+	invalidTxBz := []byte("invalid tx bytes")
+
+	tests := map[string]struct {
+		msg         *types.MsgApplyL1Txs
```
joshklop, line 34: nit: might make the test cases simpler to read if they are passed in as `[][]byte`, and then the `MsgApplyL1Txs` is built in the test itself.

x/rollup/keeper/msg_server_test.go
```diff
@@ -0,0 +1,236 @@
...
+				s.Require().Nil(resp)
+			} else {
+				s.Require().NoError(err)
+				s.Require().NotNil(resp)
+
+				// TODO: Verify that the expected event types are emitted
+
+				// Verify that the l1 block info and l1 block history are saved to the store
+				expectedBlockInfo := eth.BlockToInfo(testutils.GenerateL1Block())
+				store := sdk.UnwrapSDKContext(s.ctx).KVStore(s.storeKey)
```
joshklop, line 139: Is this how the store is retrieved in other Cosmos SDK unit tests? Is the store ever kept on the keeper suite struct?

joshklop (approved):
LGTM! Approving optimistically.

x/rollup/keeper/msg_server_test.go
natebeauregard: I was getting the following IDE warning and figured I'd remove the shadowed declaration and just reassign the previously initialized err field: `Declaration of 'err' shadows declaration at msg_server_test.go`

It's kind of unfortunate you can't conditionally initialize vars in go. Lmk if you feel strongly about changing this, otherwise I'll leave it in

x/rollup/keeper/msg_server_test.go
natebeauregard: Most cosmos sdk unit tests just use pre-existing query endpoints instead of accessing the store directly, however since we don't have any for asserting the l1 block info I opted to just check the store directly. I like your idea of keeping the store in the keeper suite so we can access it directly from the suite, I'll update that.

x/rollup/keeper/msg_server_test.go
natebeauregard: It'll be removed in the integration test [PR](https://github.com/polymerdao/monomer/pull/165/files#diff-4153058257cbc126a783b5411a175afa9a7b2a5116bd1685cae4d62526f43113R97) too since now `err` isn't declared above it :)

x/rollup/keeper/msg_server_test.go
joshklop: Once we figure out https://github.com/polymerdao/monomer/issues/160, we'll probably add query endpoints. At that point, it would be better to use the query endpoints in tests than directly accessing the store.

This works for now.

PR #166, opened 2024-08-21
Add Builder Withdrawal Tests and Refactor Related Components

### LINKED ISSUE #139: Add withdrawals tests to builder_test
Update `builder_test` to test withdrawal helper funcs and to test the full withdrawal flow when building new blocks.

We'll want to add the following tests:
- Unit test for `parseWithdrawalMessages` to assert that any withdrawal messages get the nonce attribute added to their corresponding tx result event.
- Unit test for building a block with a withdrawal message included in a tx. This will be similar to `TestBuild`, however we only care about asserting the behavior for withdrawal message types (`MsgInitiateWithdrawal`). The expected `ethStateRoot` and the expected withdrawal tx result event should be asserted in the test to verify that `builder.Build` is correctly processing withdrawal txs.

Jesserc:
This PR introduces tests for withdrawal functionality and includes some refactoring changes to improve code structure and testability.

Key changes:

1. Withdrawal Tests:
	- Added tests for withdrawal messages in `builder_test.go`
	- Implemented ToWithdrawalTx testapp helper function for creating withdrawal transactions in tests
	- Added test for parseWithdrawalMessages to verify correct parsing of withdrawal messages
	- Added test for verifying received L1 deposit transactions before withdrawal

2. Refactoring:
	- Introduced EvmToCosmos function in adapters.go for centralized EVM to Cosmos address conversion

Fixes #139 

## Summary by CodeRabbit

- **New Features**
	- Added a function to convert Ethereum addresses to Cosmos SDK account addresses, enhancing interoperability between the two platforms.
	- Introduced a utility for improved mock account retrieval functionality, allowing for more realistic testing scenarios.
	- Created a new function for initiating withdrawal transactions, expanding transaction capabilities.
	- Enhanced testing framework for withdrawal and deposit transactions, ensuring correctness throughout the transaction lifecycle.
	- Established a consistent Node.js version for the project to reduce compatibility issues.
	- Improved routing structure in documentation for better navigation.
	- Introduced a `BankKeeper` interface for consistent coin management operations within the rollup module.

- **Bug Fixes**
	- Improved error handling in account retrieval processes, enhancing mock behavior.

- **Tests**
	- Enhanced test coverage for withdrawal messages to ensure proper transaction handling and state validation.
	- Updated testing utilities for transaction creation, improving accuracy and clarity in test scenarios.

### DIFF
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -3,8 +3,10 @@ package builder_test
 import (
 	"context"
 	"fmt"
+	"math/big"
 	"testing"
 
+	"cosmossdk.io/math"
 	abcitypes "github.com/cometbft/cometbft/abci/types"
 	cmtpubsub "github.com/cometbft/cometbft/libs/pubsub"
 	bfttypes "github.com/cometbft/cometbft/types"
@@ -19,10 +21,23 @@ import (
 	"github.com/polymerdao/monomer/evm"
 	"github.com/polymerdao/monomer/genesis"
 	"github.com/polymerdao/monomer/mempool"
+	"github.com/polymerdao/monomer/monomerdb/localdb"
 	"github.com/polymerdao/monomer/testapp"
 	"github.com/polymerdao/monomer/testutils"
+	"github.com/polymerdao/monomer/utils"
+	"github.com/polymerdao/monomer/x/rollup/types"
 	"github.com/stretchr/testify/require"
 )
+
+type testEnvironment struct {
+	app        *testapp.App
+	g          *genesis.Genesis
+	pool       *mempool.Pool
+	blockStore *localdb.DB
+	txStore    txstore.TxStore
+	ethstatedb state.Database
+	eventBus   *bfttypes.EventBus
+}
 
 type queryAll struct{}
 
@@ -80,74 +95,51 @@ func TestBuild(t *testing.T) {
 
 	for description, test := range tests {
 		t.Run(description, func(t *testing.T) {
+			// +1 because we want it to be buffered even when mempool and inclusion list are empty.
+			subChannelLen := len(test.mempool) + len(test.inclusionList) + 1
+			env, subscription := setupTestEnvironmentAndSubscribe(t, subChannelLen)
+
 			inclusionListTxs := append([][]byte{testutils.GenerateBlock(t).Txs[0]}, testapp.ToTxs(t, test.inclusionList)...)
 			mempoolTxs := testapp.ToTxs(t, test.mempool)
 
-			pool := mempool.New(testutils.NewMemDB(t))
 			for _, tx := range mempoolTxs {
-				require.NoError(t, pool.Enqueue(tx))
+				require.NoError(t, env.pool.Enqueue(tx))
 			}
-			blockStore := testutils.NewLocalMemDB(t)
-			txStore := txstore.NewTxStore(testutils.NewCometMemDB(t))
-			ethstatedb := testutils.NewEthStateDB(t)
-
-			var chainID monomer.ChainID
-			app := testapp.NewTest(t, chainID.String())
-			g := &genesis.Genesis{
-				ChainID:  chainID,
-				AppState: testapp.MakeGenesisAppState(t, app),
-			}
-
-			eventBus := bfttypes.NewEventBus()
-			require.NoError(t, eventBus.Start())
-			t.Cleanup(func() {
-				require.NoError(t, eventBus.Stop())
-			})
-			// +1 because we want it to be buffered even when mempool and inclusion list are empty.
-			subChannelLen := len(test.mempool) + len(test.inclusionList) + 1
-			subscription, err := eventBus.Subscribe(context.Background(), "test", &queryAll{}, subChannelLen)
-			require.NoError(t, err)
-
-			require.NoError(t, g.Commit(context.Background(), app, blockStore, ethstatedb))
 
 			b := builder.New(
-				pool,
-				app,
-				blockStore,
-				txStore,
-				eventBus,
-				g.ChainID,
-				ethstatedb,
+				env.pool,
+				env.app,
+				env.blockStore,
+				env.txStore,
+				env.eventBus,
+				env.g.ChainID,
+				env.ethstatedb,
 			)
 
 			payload := &builder.Payload{
 				InjectedTransactions: bfttypes.ToTxs(inclusionListTxs),
 				GasLimit:             0,
-				Timestamp:            g.Time + 1,
+				Timestamp:            env.g.Time + 1,
 				NoTxPool:             test.noTxPool,
 			}
-			preBuildInfo, err := app.Info(context.Background(), &abcitypes.RequestInfo{})
-			require.NoError(t, err)
-			builtBlock, err := b.Build(context.Background(), payload)
-			require.NoError(t, err)
-			postBuildInfo, err := app.Info(context.Background(), &abcitypes.RequestInfo{})
-			require.NoError(t, err)
+
+			builtBlock, preBuildInfo, postBuildInfo := buildBlock(t, b, env.app, payload)
 
 			// Application.
 			{
 				height := uint64(postBuildInfo.GetLastBlockHeight())
-				app.StateContains(t, height, test.inclusionList)
+				env.app.StateContains(t, height, test.inclusionList)
 				if test.noTxPool {
-					app.StateDoesNotContain(t, height, test.mempool)
+					env.app.StateDoesNotContain(t, height, test.mempool)
 				} else {
-					app.StateContains(t, height, test.mempool)
+					env.app.StateContains(t, height, test.mempool)
 				}
 			}
 
 			// Block store.
-			genesisHeader, err := blockStore.BlockByHeight(uint64(preBuildInfo.GetLastBlockHeight()))
+			genesisHeader, err := env.blockStore.BlockByHeight(uint64(preBuildInfo.GetLastBlockHeight()))
 			require.NoError(t, err)
-			gotBlock, err := blockStore.HeadBlock()
+			gotBlock, err := env.blockStore.HeadBlock()
 			require.NoError(t, err)
 			allTxs := append([][]byte{}, inclusionListTxs...)
 			if !test.noTxPool {
@@ -156,7 +148,7 @@ func TestBuild(t *testing.T) {
 
 			ethStateRoot := gotBlock.Header.StateRoot
 			header := &monomer.Header{
-				ChainID:    g.ChainID,
+				ChainID:    env.g.ChainID,
 				Height:     uint64(postBuildInfo.GetLastBlockHeight()),
 				Time:       payload.Timestamp,
 				ParentHash: genesisHeader.Header.Hash,
@@ -165,11 +157,10 @@ func TestBuild(t *testing.T) {
 			}
 			wantBlock, err := monomer.MakeBlock(header, bfttypes.ToTxs(allTxs))
 			require.NoError(t, err)
-			require.Equal(t, wantBlock, builtBlock)
-			require.Equal(t, wantBlock, gotBlock)
+			verifyBlockContent(t, wantBlock, gotBlock, builtBlock)
 
 			// Eth state db.
-			ethState, err := state.New(ethStateRoot, ethstatedb, nil)
+			ethState, err := state.New(ethStateRoot, env.ethstatedb, nil)
 			require.NoError(t, err)
 			appHash, err := getAppHashFromEVM(ethState, header)
 			require.NoError(t, err)
@@ -177,31 +168,14 @@ func TestBuild(t *testing.T) {
 
 			// Tx store and event bus.
 			for i, tx := range wantBlock.Txs {
-				checkTxResult := func(got abcitypes.TxResult) {
-					// We don't check the full result, which would be difficult and a bit overkill.
-					// We only verify that the main info is correct.
-					require.Equal(t, uint32(i), got.Index)
-					require.Equal(t, wantBlock.Header.Height, uint64(got.Height))
-					require.Equal(t, tx, bfttypes.Tx(got.Tx))
-				}
-
 				// Tx store.
-				got, err := txStore.Get(tx.Hash())
+				got, err := env.txStore.Get(tx.Hash())
 				require.NoError(t, err)
-				checkTxResult(*got)
+				checkTxResult(t, got, wantBlock, i, tx)
 
 				// Event bus.
-				select {
-				case event, ok := <-subscription.Out():
-					if !ok {
-						require.FailNow(t, "event channel closed unexpectedly")
-					}
-					data := event.Data()
-					require.IsType(t, bfttypes.EventDataTx{}, data)
-					checkTxResult(data.(bfttypes.EventDataTx).TxResult)
-				case <-subscription.Canceled():
-					require.FailNow(t, "subscription channel closed unexpectedly")
-				}
+				eventData := getEventData(t, subscription)
+				checkTxResult(t, &eventData.TxResult, wantBlock, i, tx)
 			}
 			require.NoError(t, subscription.Err())
 		})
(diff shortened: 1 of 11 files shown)


builder/builder_test.go
```diff
@@ -307,3 +317,219 @@ func getAppHashFromEVM(ethState *state.StateDB, header *monomer.Header) (common.
...
+	require.NoError(t, err)
+	mockAcc := client.Account(authtypes.NewBaseAccount(s.AccountAddress(), privKey.PubKey(), 0, 0))
+	// Create a mock AccountRetriever
+	// The client context needs this for the signer
+	mockAccountRetriever := e2e.MockAccountRetriever{
+		Accounts: map[string]*client.Account{
+			s.AccountAddress().String(): &mockAcc,
+		},
+	}
+	clientCtx.AccountRetriever = mockAccountRetriever
```
natebeauregard, line 377: This may resolve a lot of my other comments so feel free to ignore them if you implement this. Instead of building all of this signer info, could we just use `GenerateBlock` like the builder test does to create the block transactions? We shouldn't need to use a signer since we're never actually executing the txs in the appchain, we're only testing that the block can be properly built.

natebeauregard (commented):
Started my review of the test setup, I still need to review the actual test and the deposit/withdrawal assertions. My most important comment relates to how we're building the txs and to see if we could just use `GenerateBlock` since we're never actually running these txs in the L2 appchain. I think a lot of the signer setup can just be omitted.

builder/builder_test.go
Jesserc: Fixed

builder/builder_test.go
```diff
@@ -307,3 +317,219 @@ func getAppHashFromEVM(ethState *state.StateDB, header *monomer.Header) (common.
...
+
+	depositTx, err := monomer.AdaptPayloadTxsToCosmosTxs([]hexutil.Bytes{l1InfoTxBytes, depositTxBytes}, s.Sign, s.AccountAddress().String())
+	require.NoError(t, err)
+
+	cosmAddr := monomer.EvmToCosmos(*depositTxETH.To()) // The recipient of the deposit, will initiate the withdrawal
+	ethAddr := common.HexToAddress("0x12345abcde")
+	withdrawalAmount := depositTxETH.Value()
```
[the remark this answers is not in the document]
Jesserc: `cosmAddr` is still used after this line, but I've inlined `ethAddr`

builder/builder_test.go
```diff
@@ -36,6 +38,73 @@ func (*queryAll) String() string {
...
+	t.Cleanup(func() {
+		require.NoError(t, eventBus.Stop())
+	})
+
+	require.NoError(t, g.Commit(context.Background(), app, blockStore, ethstatedb))
+
+	return app, g, pool, blockStore, txStore, ethstatedb, eventBus
+}
+
+func createBuilder(pool *mempool.Pool, app *testapp.App, blockStore *localdb.DB, txStore txstore.TxStore, eventBus *bfttypes.EventBus, chainID monomer.ChainID, ethstatedb state.Database) *builder.Builder {
```
natebeauregard, line 66: I feel like it would be cleaner to remove the `createBuilder` func and just inline `builder.New(...)` in the tests

builder/builder_test.go
```diff
@@ -36,6 +38,73 @@ func (*queryAll) String() string {
 	return "all"
 }
 
+//nolint:gocritic // tooManyResultsChecker: this function needs to return multiple values for test setup
+func setupTestEnvironment(t *testing.T) (*testapp.App, *genesis.Genesis, *mempool.Pool, *localdb.DB, txstore.TxStore, state.Database, *bfttypes.EventBus) {
```
natebeauregard, line 42: It may be better to just create a `testEnvironment` struct instead of needing to suppress the lint warning.

Also, can we use the test environment in the `TestRollback` test?

builder/builder_test.go
```diff
@@ -307,3 +339,175 @@ func getAppHashFromEVM(ethState *state.StateDB, header *monomer.Header) (common.
...
+	require.Equal(t, expectedStateRoot, gotStateRoot, "Expected the built block to contain state root hash")
+
+	for i := range gotBlock.Txs {
+		select {
+		case event, ok := <-subscription.Out():
+			if !ok {
+				require.FailNow(t, "event channel closed unexpectedly")
+			}
+			data := event.Data()
+			require.IsType(t, bfttypes.EventDataTx{}, data)
```
natebeauregard, line 431: This check isn't necessary due to the type assertion below

```suggestion
```

builder/builder_test.go
```diff
@@ -307,3 +339,175 @@ func getAppHashFromEVM(ethState *state.StateDB, header *monomer.Header) (common.
...
+			e, ok := data.(bfttypes.EventDataTx)
+			require.True(t, ok)
+			// Deposit transaction
+			if i == 0 {
+				require.Equal(t, e.GetResult().Events[0].Attributes[0].Value, "/rollup.v1.MsgApplyL1Txs")
+			}
+			// Withdrawal transaction
+			if i == 1 {
+				require.Equal(t, e.GetResult().Events[0].Attributes[0].Value, "/rollup.v1.MsgInitiateWithdrawal")
+			}
```
natebeauregard, line 441: Instead of using indexes to differentiate between the deposit and withdrawal txs, could you check the event type instead to differentiate? That seems cleaner to me and will help with readability

builder/builder_test.go
```diff
@@ -307,3 +339,175 @@ func getAppHashFromEVM(ethState *state.StateDB, header *monomer.Header) (common.
...
+		}
+		require.NoError(t, subscription.Err())
+	}
+}
+
+func testParseWithdrawalMessages(t *testing.T, withdrawalTxResult *abcitypes.TxResult) {
+	const (
+		nonceKey          = "nonce"
+		expectedEventType = "withdrawal_initiated"
+	)
```
natebeauregard, line 454: Instead of creating new test constants, could you use the actual constants from `x/rollup/types/events.go` for these values?

`AttributeKeyNonce`
`EventTypeWithdrawalInitiated`

builder/builder_test.go
```diff
@@ -307,3 +339,175 @@ func getAppHashFromEVM(ethState *state.StateDB, header *monomer.Header) (common.
...
+	require.NotEmpty(t, nonceAttribute.Value, "Withdrawal nonce value should not be empty")
+}
+
+func checkDepositTXResult(t *testing.T, txStore txstore.TxStore, depositTx bfttypes.Tx, expectedDepositAmount, cosmAddr string) {
+	const (
+		eventType              = "coin_received"
+		receiverAttributeIndex = 0
+		valueAttributeIndex    = 1
+		targetEventCount       = 2
+	)
```
natebeauregard, line 487: Instead of using the bank modules `coin_received` event, could we use the `x/rollup` module's `EventTypeMintETH` event and inline it instead of creating a new constant here?

builder/builder_test.go
```diff
@@ -307,3 +308,211 @@ func getAppHashFromEVM(ethState *state.StateDB, header *monomer.Header) (common.
...
+
+	for i, event := range depositTxResult.Result.Events {
+		if event.Type == eventType {
+			eventCount++
+			if eventCount == targetEventCount {
+				coinReceivedEvent = &depositTxResult.Result.Events[i]
+				break
+			}
+		}
+	}
```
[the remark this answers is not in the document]
Jesserc: @natebeauregard The `coin_received` event type appears twice; I think the first is when the coin is minted, and the second is when it is finally sent to our target. The receiver in the first one isn't our target. That's why I had a target count.

This is what the event looks like.

```json
[
  {
    "message": [
      {"action": "/rollup.v1.MsgApplyL1Txs", "true": true},
      {"sender": "cosmos1v9jxguspv4h2u", "true": true},
      {"msg_index": 0, "true": true}
    ]
  },
  {
    "coin_received": [
      {"receiver": "cosmos1ghacknt8xydqjp0ghnl7rad8966wv272hlde75", "true": true},
      {"amount": "193000000000000000000ETH", "true": true},
      {"msg_index": 0, "true": true}
    ]
  },
  {
    "coinbase": [
      {"minter": "cosmos1ghacknt8xydqjp0ghnl7rad8966wv272hlde75", "true": true},
      {"amount": "193000000000000000000ETH", "true": true},
      {"msg_index": 0, "true": true}
    ]
  },
  {
    "coin_spent": [
      {"spender": "cosmos1ghacknt8xydqjp0ghnl7rad8966wv272hlde75", "true": true},
      {"amount": "193000000000000000000ETH", "true": true},
      {"msg_index": 0, "true": true}
    ]
  },
  {
    "coin_received": [
      {"receiver": "cosmos1wmufz6m7jjekmk7kafqvpeaveg4gx993u9qvsf", "true": true},
      {"amount": "193000000000000000000ETH", "true": true},
      {"msg_index": 0, "true": true}
    ]
  },
  {
    "transfer": [
      {"recipient": "cosmos1wmufz6m7jjekmk7kafqvpeaveg4gx993u9qvsf", "true": true},
      {"sender": "cosmos1ghacknt8xydqjp0ghnl7rad8966wv272hlde75", "true": true},
      {"amount": "193000000000000000000ETH", "true": true},
      {"msg_index": 0, "true": true}
    ]
  },
  {
    "message": [
      {"sender": "cosmos1ghacknt8xydqjp0ghnl7rad8966wv272hlde75", "true": true},
      {"msg_index": 0, "true": true}
    ]
  },
  {
    "message": [
      {"module": "rollup", "true": true},
      {"msg_index": 0, "true": true}
    ]
  },
  {
    "mint_eth": [
      {"l1_deposit_tx_type": "l1_user_deposit", "true": true},
      {"to_cosmos_address": "cosmos1wmufz6m7jjekmk7kafqvpeaveg4gx993u9qvsf", "true": true},
      {"value": "0x0a7669bd7132640000", "true": true},
      {"msg_index": 0, "true": true}
    ]
  }
]
```

builder/builder_test.go
Jesserc: @natebeauregard Their types are both "message"

```
{message [{action /rollup.v1.MsgApplyL1Txs true} {sender cosmos1v9jxguspv4h2u true} {msg_index 0 true}]}

{message [{action /rollup.v1.MsgInitiateWithdrawal true} {sender cosmos1wmufz6m7jjekmk7kafqvpeaveg4gx993u9qvsf true} {msg_index 0 true}]}
```

I'll go ahead and use the type instead of index. Thanks

Jesserc: @natebeauregard on second glance, this will cause an error since both if statements will hit on first iteration

adapters.go
```diff
@@ -132,3 +134,8 @@ func AdaptNonDepositCosmosTxToEthTx(cosmosTx bfttypes.Tx) *ethtypes.Transaction
 		Data: cosmosTx,
 	})
 }
+
+// EvmToCosmos converts an EVM address to a sdk.AccAddress
+func EvmToCosmos(addr common.Address) sdktypes.AccAddress {
```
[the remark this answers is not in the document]
joshklop: Should this be in the `monomer` package?

builder/builder_test.go
```diff
@@ -36,6 +39,61 @@ func (*queryAll) String() string {
 	return "all"
 }
 
+//nolint:gocritic // tooManyResultsChecker: this function needs to return multiple values for test setup
+func setupTestEnvironment(t *testing.T) (*testapp.App, *genesis.Genesis, *mempool.Pool, *localdb.DB, txstore.TxStore, state.Database, *bfttypes.EventBus) {
+	var chainID monomer.ChainID
```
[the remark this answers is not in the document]
Jesserc: 😅We're using the default value, you 🐰🐰

builder/builder_test.go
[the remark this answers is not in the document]
Jesserc: This has been resolved, while resolving [this](https://github.com/polymerdao/monomer/pull/166#discussion_r1729330084)

builder/builder_test.go
```diff
@@ -80,35 +161,26 @@ func TestBuild(t *testing.T) {
...
 		t.Run(description, func(t *testing.T) {
+			env := setupTestEnvironment(t)
+
+			app := env.app
+			g := env.g
+			pool := env.pool
+			blockStore := env.blockStore
+			txStore := env.txStore
+			ethstatedb := env.ethstatedb
+			eventBus := env.eventBus
```
natebeauregard, line 172: Rather than pulling each field off of the `testEnvironment` struct and storing them as vars after setup, it may be cleaner to just pull them directly off of `testEnvironment` whenever needed. This comment would also apply for the other cases where we call `setupTestEnvironment`

builder/builder_test.go
```diff
@@ -307,3 +358,178 @@ func getAppHashFromEVM(ethState *state.StateDB, header *monomer.Header) (common.
...
+	// Test parseWithdrawalMessages
+	checkWithdrawalTxResult(t, withdrawalTxResult)
+
+	expectedStateRoot := wantBlock.Header.StateRoot
+	gotStateRoot := gotBlock.Header.StateRoot
+	require.Equal(t, expectedStateRoot, gotStateRoot, "Expected the built block to contain state root hash")
+
+	for i := range gotBlock.Txs {
+		select {
+		case event, ok := <-subscription.Out():
```
natebeauregard, line 459: Could you extract the `select` statement out into a helper func since it's the same logic that the `TestBuild` test uses to verify the event bus is behaving properly?

builder/builder_test.go
natebeauregard: What if you instead just extracted out the `select` statement into a helper func as suggested in my comment above? Then you should be able to loop over the deposit/withdrawal expected event attributes without needing to rely on indexing? I feel like that would be cleaner than using indexes to verify the expected events

natebeauregard (commented):
Looking really good! Left a few more suggestions and responses to get this over the finish line

adapters.go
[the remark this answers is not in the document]
natebeauregard: @Jesserc Could you respond to Josh's question above?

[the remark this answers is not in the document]
Jesserc: @joshklop It might be appropriate to place it somewhere else, maybe `utils/utils.go`?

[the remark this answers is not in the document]
Jesserc: I moved it to `utils/utils.go`

x/rollup/keeper/deposits.go
```diff
@@ -87,7 +87,7 @@ func (k *Keeper) processL1UserDepositTxs(ctx sdk.Context, txs [][]byte) error {
 			ctx.Logger().Error("Contract creation txs are not supported", "index", i)
 			return types.WrapError(types.ErrInvalidL1Txs, "Contract creation txs are not supported, index:%d", i)
 		}
-		cosmAddr := evmToCosmos(*to)
+		cosmAddr := utils.EvmToCosmosAddress(*to)
```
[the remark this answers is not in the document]
Jesserc: The function doesn't return an error
There's no `types.AddrLen`

builder/builder_test.go
Jesserc: Done

builder/builder_test.go
```diff
@@ -307,3 +341,170 @@
...
+	require.NotNil(t, depositTxETH.Value())
+	require.NotNil(t, depositTxETH.To(), "Deposit transaction must have a 'to' address")
+
+	cosmAddr := utils.EvmToCosmosAddress(*depositTxETH.To())
+	withdrawalTx := testapp.ToWithdrawalTx(
+		t,
+		cosmAddr.String(),
+		common.HexToAddress("0x12345abcde").String(),
+		math.NewIntFromBigInt(depositTxETH.Value()),
+	)
```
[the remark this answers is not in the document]
Jesserc: `cosmAddr` is also used in checkDepositTxResult helper func and called in TestWithdrawal

builder/builder_test.go
```diff
@@ -508,3 +486,16 @@ func checkDepositTxResult(t *testing.T, txStore txstore.TxStore, depositTx bftty
 	require.Equal(t, cosmAddr, actualReceiver, "Receiver address mismatch")
 	require.Equal(t, expectedDepositAmount, actualDepositAmount, "Deposit amount mismatch")
 }
+
+func checkEventBusEventAndGetData(t *testing.T, subscription bfttypes.Subscription) bfttypes.EventDataTx {
```
natebeauregard, line 492: nit: could you rename this to either `getEventData` or `getEventDataTx`?

builder/builder_test.go
```diff
@@ -333,12 +338,9 @@ func getAppHashFromEVM(ethState *state.StateDB, header *monomer.Header) (common.
 }
 
 func TestWithdrawal(t *testing.T) {
```
natebeauregard, line 340: nit: Could `TestWithdrawal` be renamed to `TestBuildRollupTxs` and moved below the `TestBuild` test for readability?

natebeauregard (approved):
LGTM! I left two more nits that should be addressed before pushing the separate commits to get the PR ready for merge, however I trust that they'll be done and will review them in the final commit structure.

In terms of commit structure, I think this would make sense to split up into three distinct commits:
1. `ToTx` -> `ToTestTx` rename
2. moving address conversion to `utils.go`
3. `builder_test` updates

What do you think about that commit structure? Feel free to chime in if you think that there are other chunks that should be separated.

builder/builder_test.go
Jesserc: Resolved

Jesserc:
> In terms of commit structure, I think this would make sense to split up into three distinct commits:
> 
> 1. `ToTx` -> `ToTestTx` rename
> 2. moving address conversion to `utils.go`
> 3. `builder_test` updates
> 
> What do you think about that commit structure? Feel free to chime in if you think that there are other chunks that should be separated.

@natebeauregard Sounds good to me

builder/builder_test.go
```diff
@@ -36,6 +51,79 @@ func (*queryAll) String() string {
 	return "all"
 }
 
+func setupTestEnvironment(t *testing.T) testEnvironment {
```
natebeauregard, line 54: Can you move the new test helpers below the actual tests (`TestBuild`, `TestRollback`, etc) for readability?

utils/utils.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/cosmos/cosmos-sdk/types"
```
natebeauregard, line 7: Can you use the `sdktypes` or `sdk` package alias here and then update the comment above the `EvmToCosmosAddress` to reflect the accurate package alias?

natebeauregard (commented):
Couple more tiny fixes, I'll merge after they're addressed. Feel free to force push to include the changes in the commits that introduced them

Jesserc:
@natebeauregard LGTM!

natebeauregard (approved):
LGTM

PR #174, opened 2024-08-27
Emit single `message` event when processing deposits

### LINKED ISSUE #162: Only emit one sdk.EventTypeMessage for MsgApplyL1Txs
Currently, all event emission for `MsgApplyL1Txs` occurs in `mintETH` in [`deposits.go`](https://github.com/polymerdao/monomer/blob/main/x/rollup/keeper/deposits.go#L109-L120). The `sdk.EventTypeMessage` event type is emitted for each `mintETH` call instead of once for the entire message. 

We'll want to update `deposits.go` to only emit each individual mint event and update `msg_server.go` to emit the message event. Also, a unit test assertion similar to the [withdrawals event assertion](https://github.com/polymerdao/monomer/blob/b935190/x/rollup/keeper/msg_server_test.go#L184-L192) should be added in place of the [TODO](https://github.com/polymerdao/monomer/blob/b935190/x/rollup/keeper/msg_server_test.go#L108) in `msg_server_test.go` to verify the correct events are emitted.

NiloCK:
PR closes #162.

PR 
- alters `deposits.go/processL1UserDepositTxs` to gather and _return_ (instead of emitting) mint events _only_
- alters `msg_server.go/ApplyL1Txs` to craft the `sdk.EventTypeMessage` message, and emit along with the gathered mint events
- adds a test assertion that the correct events are emitted on well-formed collections of deposit TXs
  - adds the EventManager as part of the constructed test suite for visibility
  - adds a success test case with no user deposit TXs (`[0, 1, some]` seems more complete than `[1, some]`)

  
Drive-by question: why is it that we want to emit the `sdk.EventTypeMessage` message? Maybe a cosmos convention that modules prepend their emissions with a module-scope?

## Summary by CodeRabbit

- **New Features**
  - Enhanced user deposit transaction processing with improved event handling and error reporting.
  - Streamlined feedback on ETH minting events, providing users with clearer insights into transaction outcomes.
  - Introduced a new function for emitting structured events related to transactions, improving event tracking.

- **Bug Fixes**
  - Improved error handling in transaction processing to ensure events are only emitted on successful transactions.

- **Tests**
  - Updated test cases to verify expected event types during transaction processing, enhancing the reliability of the test suite.

### DIFF
--- a/x/rollup/keeper/deposits.go
+++ b/x/rollup/keeper/deposits.go
@@ -63,8 +63,8 @@ func (k *Keeper) processL1AttributesTx(ctx sdk.Context, txBytes []byte) (*derive
 }
 
 // processL1UserDepositTxs processes the L1 user deposit txs, mints ETH to the user's cosmos address,
-// and emits the minted ETH events.
-func (k *Keeper) processL1UserDepositTxs(ctx sdk.Context, txs [][]byte) error { //nolint:gocritic // hugeParam
+// and returns associated events.
+func (k *Keeper) processL1UserDepositTxs(ctx sdk.Context, txs [][]byte) (sdk.Events, error) { //nolint:gocritic // hugeParam
 	events := sdk.Events{}
 
 	for i := 1; i < len(txs); i++ {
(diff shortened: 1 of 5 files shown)


natebeauregard:
> Drive-by question: why is it that we want to emit the sdk.EventTypeMessage message? Maybe a cosmos convention that modules prepend their emissions with a module-scope?

@NiloCK The `sdk.EventTypeMessage` message convention was inherited by the original monomer implementation for deposits so I just decided to keep that convention for withdrawals. This is a good reminder that we should probably make an issue to review and evaluate whether our event types/attributes are valuable and not redundant. I think some usability testing for events would be good to do before release.

natebeauregard (approved):
:rocket:

PR #175, opened 2024-08-27
[Optional] Add an EmitEvents helper to de-duplicate message event construction

NiloCK:
Currently, `msg_server.go` emits events in two locations. In each location, it constructs a `message` event with its module name attached:

```go
sdk.NewEvent(
	sdk.EventTypeMessage,
	sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
)
```

This is somewhat error-prone - could forget to prepend these messages on other event emissions.

This PR adds a helper to the `keeper` struct to encapsulate this `message` event prepend operation.

### DIFF
--- a/x/rollup/keeper/keeper.go
+++ b/x/rollup/keeper/keeper.go
@@ -1,8 +1,11 @@
 package keeper
 
 import (
+	"context"
+
 	"cosmossdk.io/core/store"
 	"github.com/cosmos/cosmos-sdk/codec"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/ethereum-optimism/optimism/op-node/rollup"
 	"github.com/polymerdao/monomer/x/rollup/types"
 )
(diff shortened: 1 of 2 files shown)


x/rollup/keeper/keeper.go
```diff
@@ -27,3 +30,17 @@ func NewKeeper(
 		rollupCfg:    &rollup.Config{},
 	}
 }
+
+// Helper. Prepares a `message` event with the module name and emits it
+// along with the provided events.
+func (k *Keeper) EmitEvents(goCtx context.Context, events sdk.Events) {
```
natebeauregard, line 36: It may be worth naming `EmitEvents` to something like `EmitMessageEvents`, `EmitRollupEvents`, or `EmitModuleEvents` to avoid confusion with the event manager `EmitEvents` func.

natebeauregard (approved):

x/rollup/keeper/keeper.go
NiloCK: There are a few spots in the codebase where overlapping function names are included, the idea being that the package / struct that they are attached to effectively namespaces them.

I think I'll leave this as-is for now, but happy to revise in future if it provokes any confusion.

PR #179, opened 2024-08-28
Remove `msgServer` type

### LINKED ISSUE #167: Remove `msgServer` type if possible
https://github.com/polymerdao/monomer/blob/13157874d7fb51091a77a4fa9ac8904730422c7f/x/rollup/keeper/msg_server.go#L13-L23

AnkushinDaniil:
Fixes #167

## Summary by CodeRabbit

- **New Features**
	- Streamlined message server registration process, enhancing performance and clarity.

- **Bug Fixes**
	- Removed unnecessary abstractions in message server implementation, improving code maintainability.

- **Tests**
	- Simplified test logic by directly invoking methods on the rollupKeeper instance, enhancing readability.

- **Documentation**
	- Improved code readability and maintainability through adjustments in import statements and function calls.

### DIFF
--- a/x/rollup/keeper/msg_server.go
+++ b/x/rollup/keeper/msg_server.go
@@ -10,6 +10,8 @@ import (
 	"github.com/samber/lo"
 )
 
+var _ types.MsgServer = &Keeper{}
+
 // ApplyL1Txs implements types.MsgServer.
 func (k *Keeper) ApplyL1Txs(goCtx context.Context, msg *types.MsgApplyL1Txs) (*types.MsgApplyL1TxsResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
--- a/x/rollup/keeper/msg_server.go
+++ b/x/rollup/keeper/msg_server.go
@@ -10,18 +10,6 @@ import (
 	"github.com/samber/lo"
 )
 
-type msgServer struct {
-	*Keeper
-}
-
-// NewMsgServerImpl returns an implementation of the MsgServer interface
-// for the provided Keeper.
-func NewMsgServerImpl(keeper *Keeper) types.MsgServer {
-	return &msgServer{Keeper: keeper}
-}
-
-var _ types.MsgServer = msgServer{}
-
 // ApplyL1Txs implements types.MsgServer.
 func (k *Keeper) ApplyL1Txs(goCtx context.Context, msg *types.MsgApplyL1Txs) (*types.MsgApplyL1TxsResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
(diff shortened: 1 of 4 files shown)


AnkushinDaniil:
@joshklop

AnkushinDaniil:
@natebeauregard

x/rollup/keeper/msg_server.go
```diff
@@ -10,18 +10,6 @@ import (
...
-	*Keeper
-}
-
-// NewMsgServerImpl returns an implementation of the MsgServer interface
-// for the provided Keeper.
-func NewMsgServerImpl(keeper *Keeper) types.MsgServer {
-	return &msgServer{Keeper: keeper}
-}
-
-var _ types.MsgServer = msgServer{}
```
natebeauregard, line 23: We may still want to leave in the compile time interface check here, it seems like a common pattern for cosmos sdk msg servers.

`var _ types.MsgServer = &Keeper{}`

x/rollup/keeper/msg_server.go
natebeauregard, line 1: After looking at a few different Cosmos SDK modules it seems like they're all using the `msgServer` struct wrapper instead of having the keeper implement the interface directly. However, the [Cosmos SDK docs](https://docs.cosmos.network/main/build/building-modules/msg-services#implementation-of-a-module-msg-service) mention that both approaches are fine, which a slight preference to having the keeper implement the `MsgServer` interface directly:

> When possible, the existing module's [Keeper](https://docs.cosmos.network/main/build/building-modules/keeper) should implement MsgServer, otherwise a msgServer struct that embeds the Keeper can be created, typically in ./keeper/msg_server.go

It's kind of interesting to me that none of the standard Cosmos SDK modules have the keeper implement `MsgServer` directly however, were you able to find any precedent for modules that have their msg server set up this way? I'd be fine with merging this as is since I feel that it's a cleaner approach, however I worry a little bit about consistency with other cosmos sdk modules.

Tagging @joshklop as well for visibility - feel free to chime in with any opinions or additional info

natebeauregard (commented):
Left a couple small questions. I'll approve after they're addressed or resolved

x/rollup/keeper/msg_server.go
joshklop: I think removing it is preferable. Keeping it doesn't make sense to me.

Thanks for looking up the conventions and writing up the summary.

x/rollup/keeper/msg_server.go
AnkushinDaniil: [Add interface check](https://github.com/polymerdao/monomer/pull/179/commits/6aefe17933ec159ec455986946a5f07b89cf3799)

natebeauregard (approved):

PR #185, opened 2024-08-30
Refactor e2e tests: Replace static blob artifact dir with `os.TempDir()`

### LINKED ISSUE #173: Remove `blobs_bundle_1.json` e2e artifact
The geth node used in e2e tests produces an empty log artifact `blobs_bundle_1.json`.

Since we are not using blobs, it'd be better not to write this.

See https://github.com/polymerdao/monomer/blob/1ae2e9dec770b5cca811e1997cb8078c351f55ad/e2e/gethdevnet.go#L17 - probably better to write to a system temp directory via `os.TempDir()`.

Jesserc:
Fixes #173

This PR addresses the issue of an unused `blobs_bundle_1.json` artifact being generated during e2e tests. Since Monomer is not currently using blobs, this file is unnecessary.

Changes:
- Refactored `gethdevnet.go` to use `os.TempDir()` instead of a hardcoded artifacts directory for blobs

## Summary by CodeRabbit

- **New Features**
	- Updated blob storage management to utilize the system's temporary directory, enhancing file management and cleanup processes during execution.

### DIFF
--- a/e2e/gethdevnet.go
+++ b/e2e/gethdevnet.go
@@ -2,7 +2,7 @@ package e2e
 
 import (
 	"fmt"
-	"path/filepath"
+	"os"
 	"time"
 
 	"github.com/ethereum-optimism/optimism/op-e2e/e2eutils/fakebeacon"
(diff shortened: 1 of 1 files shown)


NiloCK (approved):
:+1: 

The blobs are no longer present in the e2e test artifact.

PR #191, opened 2024-09-04
Fix e2e test rollbacks

joshklop:
Derivation will fail [here] if the sequencer does not post with the same address as the one specified in the onchain `SystemConfig`. Unfortunately, this cannot be counted as a fatal error during derivation since anyone can technically post batcher transactions. As a result, derivation will silently continue with empty batches, which manifests itself in rollbacks every time unsafe block consolidation is attempted when the sequencing window expires.

The `SystemConfig` is set when we convert the `DeployConfig` to a `RollupConfig`. Thus, to fix the issue, we set the `BatcherSenderAddress` in the `DeployConfig`.

[here]: https://github.com/ethereum-optimism/optimism/blob/24a8d3e06e61c7a8938dfb7a591345a437036381/op-node/rollup/derive/data_source.go#L110-L114

## Summary by CodeRabbit

- **New Features**
	- Enhanced configuration for deployment settings to optimize performance during testing.
	- Introduced a new assignment for the batch sender's address, improving transaction processing.

- **Bug Fixes**
	- Adjusted logic for handling sequencer window size to facilitate unsafe block consolidation during tests.

- **Refactor**
	- Removed outdated settings related to channel timeout, streamlining the batching process.

### DIFF
--- a/e2e/op_stack.go
+++ b/e2e/op_stack.go
@@ -228,10 +228,9 @@ func (op *OPStack) runBatcher(ctx context.Context, env *environment.Env, l1Clien
 		L1Client:         l1Client,
 		EndpointProvider: endpointProvider,
 		ChannelConfig: batcher.ChannelConfig{
-			SeqWindowSize:      op.rollupConfig.SeqWindowSize,
-			ChannelTimeout:     op.rollupConfig.ChannelTimeout,
-			MaxChannelDuration: 1,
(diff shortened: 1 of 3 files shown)


natebeauregard (approved):
LGTM :rocket:

natebeauregard (approved):

Issue #189, opened 2024-09-04
Add x/rollup query endpoint for getting L1BlockInfo


natebeauregard:
Expose a new query endpoint for accessing the latest L1BlockInfo from the `x/rollup` store. The approach used for the [previous L1BlockInfo query endpoint](https://github.com/polymerdao/monomer/pull/78) can be used for guidance.

[Cosmos SDK Query Docs](https://docs.cosmos.network/main/build/building-modules/messages-and-queries#queries)

natebeauregard:
This issue will be incorporated in #291

PR #192, opened 2024-09-04
Add e2e test case to check for monomer rollbacks

natebeauregard:
Checks to ensure that the latest monomer block height is never less than the last checked monomer block height.


e2e/stack_test.go
```diff
@@ -117,6 +122,26 @@ func TestE2E(t *testing.T) {
...
+	require.NoError(t, err)
+	for {
+		currentBlock, err := stack.MonomerClient.BlockByNumber(stack.Ctx, nil)
+		require.NoError(t, err)
+		// Ensure that the current monomer block height isn't less than the last checked monomer block height.
+		if currentBlock.Number().Int64() < lastBlock.Number().Int64() {
+			require.Fail(t, "monomer has rolled back")
+		}
+		// End the test once the target monomer block height is reached
+		if currentBlock.Number().Int64() >= int64(15) {
```
natebeauregard, line 136: question for reviewers - With a SWS of 4, L1 block time of 2 ,and L2 block time of 1, the rollbacks will always occur at monomer block 13 so this check is sufficient. However, if any config param ever changes, the target monomer block height may need to be tuned to correctly check for rollbacks. Should I make calculating the target block height more formulaic based on those config params?

joshklop: We should hit the rollback after a single sequencing window has passed on L1, correct? Therefore, it makes sense to wait one SWS + 1 L1 blocks.

e2e/stack_test.go
```diff
@@ -117,6 +122,46 @@ func TestE2E(t *testing.T) {
...
+			currentHeight := eventNewBlockHeader.Header.Height
+
+			// Ensure that the current block height is not less than the last checked height
+			if lastBlockHeight > 0 && currentHeight < lastBlockHeight {
+				require.FailNow(t, "monomer has rolled back")
+			}
+			lastBlockHeight = currentHeight
+
+		// End the test once a sequencing epoch has passed.
+		case <-l1BlockTicker.C:
```
joshklop, line 154: Rather than using a ticker, can we check the L1Info on the block to determine the epoch? This also means we won't need to add the `SequencingWindowSize` to the `StackConfig` struct.

e2e/stack_test.go
```diff
@@ -117,6 +122,46 @@ func TestE2E(t *testing.T) {
...
+		select {
+		// Check new block header events for rollbacks
+		case event := <-eventChan:
+			eventNewBlockHeader, ok := event.Data.(bfttypes.EventDataNewBlockHeader)
+			require.True(t, ok)
+			currentHeight := eventNewBlockHeader.Header.Height
+
+			// Ensure that the current block height is not less than the last checked height
+			if lastBlockHeight > 0 && currentHeight < lastBlockHeight {
+				require.FailNow(t, "monomer has rolled back")
```
joshklop, line 149: nit: I think we can use slightly more idiomatic `require`s:

```golang
require.Equal(t, currentHeight, lastBlockHeight+1)
```

If I'm understanding correctly, we may still need to guard this by an `if` statement to properly initialize `lastBlockHeight`:

```golang
if lastBlockHeight > 0 {
    require.Equal(t, currentHeight, lastBlockHeight+1)
}
```

If that understanding is correct, it would be good to add a comment above the conditional explaining why we need `lastBlockHeight > 0`.

natebeauregard:
force-pushed to rebase on main

e2e/stack_test.go
natebeauregard: Good idea, added https://github.com/polymerdao/monomer/pull/192/commits/dd88c16a87206cc8cc9d9d14a6d3951e48913ae8

e2e/stack_test.go
natebeauregard: Are you referring to using the `x/rollup` query endpoint to get the L1 block info or just extracting the L1 info from the L1 attributes tx?

The query endpoint still isn't implemented, that work is tracked by [this issue](https://github.com/polymerdao/monomer/issues/189).

joshklop: I mean checking the first transaction in the block.

e2e/stack_test.go
```diff
@@ -117,6 +124,44 @@ func TestE2E(t *testing.T) {
...
+		require.NoError(t, err)
+		l1BlockInfo, err := derive.L1BlockInfoFromBytes(&rollup.Config{}, uint64(eventNewBlock.Block.Time.Unix()), ethTxs[0].Data())
+		require.NoError(t, err)
+
+		// End the test once a sequencing window has passed.
+		if l1BlockInfo.Number >= stack.SequencerWindowSize+1 {
+			t.Log("No Monomer rollbacks detected")
+			return
+		}
+	}
```
joshklop, line 162: We should fail the test here since a premature close on the `eventChan` is bad.

e2e/stack_test.go
```diff
@@ -117,6 +124,44 @@ func TestE2E(t *testing.T) {
...
+		if lastBlockHeight > 0 {
+			// Ensure that the current block height is the last checked height + 1
+			require.Equal(t, currentHeight, lastBlockHeight+1, "monomer has rolled back")
+		}
+		lastBlockHeight = currentHeight
+
+		// Get the L1 block info from the first tx in the block
+		ethTxs, err := monomer.AdaptCosmosTxsToEthTxs(eventNewBlock.Block.Txs)
+		require.NoError(t, err)
+		l1BlockInfo, err := derive.L1BlockInfoFromBytes(&rollup.Config{}, uint64(eventNewBlock.Block.Time.Unix()), ethTxs[0].Data())
```
joshklop, line 154: Might be best to just attach the full rollup config to the StackConfig now that I think about it. That will give us the SequencerWindowSize for free.

e2e/stack_test.go
```diff
@@ -117,6 +122,47 @@ func TestE2E(t *testing.T) {
...
+		// Check new block header events for rollbacks
+		case event := <-eventChan:
+			eventNewBlockHeader, ok := event.Data.(bfttypes.EventDataNewBlockHeader)
+			require.True(t, ok)
+			currentHeight := eventNewBlockHeader.Header.Height
+
+			// Skip the rollback check if lastBlockHeight is not initialized yet
+			if lastBlockHeight > 0 {
+				// Ensure that the current block height is the last checked height + 1
+				require.Equal(t, currentHeight, lastBlockHeight+1, "monomer has rolled back")
```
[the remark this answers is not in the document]
natebeauregard: Talked over chat - we're going to leave this with the block height check for now since Monomer doesn't convert the parent hash to cometbft headers

PR #193, opened 2024-09-06
Emit correct monomer block events

natebeauregard:
### Background
Currently, the block events emitted in `EventDataNewBlock` contains all of the tx events concatentated together. However, the CometBFT implementation uses the `ResponseFinalizeBlock.Events` field for this event attribute instead.

### Changes
Updates the block events emitted by the `EventDataNewBlock` event to be pulled directly from `ResponseFinalizeBlock`.
Additionally, adds some more test assertions for the fields pulled from `ResponseFinalizeBlock`.

## Summary by CodeRabbit

- **New Features**
	- Improved event publishing process by directly utilizing response events, streamlining event handling and potentially enhancing performance.

- **Bug Fixes**
	- Enhanced test clarity and specificity regarding transaction results and event generation, ensuring accurate verification of expected outcomes.

### DIFF
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -168,7 +168,7 @@ func TestBuild(t *testing.T) {
 			require.Equal(t, appHash[:], postBuildInfo.GetLastBlockAppHash())
 
 			// Tx store and event bus.
-			wantBlockEvents := make([]abcitypes.Event, 0, len(wantBlock.Txs))
+			expectedTxResults := make([]*abcitypes.ExecTxResult, 0, len(wantBlock.Txs))
 			for i, tx := range wantBlock.Txs {
 				// Tx store.
 				got, err := env.txStore.Get(tx.Hash())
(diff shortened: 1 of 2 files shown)


builder/builder_test.go
```diff
@@ -179,17 +179,21 @@ func TestBuild(t *testing.T) {
...
 			require.Equal(t, int64(len(wantBlock.Txs)), eventDataNewBlockEvents.NumTxs)
-			require.Equal(t, wantBlockEvents, eventDataNewBlockEvents.Events)
+			require.Equal(t, expectedBlockEvents, eventDataNewBlockEvents.Events)
 
 			eventDataNewBlock := getEventData[bfttypes.EventDataNewBlock](t, subscription)
 			require.Equal(t, wantBlock.ToCometLikeBlock(), eventDataNewBlock.Block)
 			require.Equal(t, bytes.HexBytes(wantBlock.Header.Hash.Bytes()), eventDataNewBlock.BlockID.Hash)
+			require.Equal(t, expectedBlockEvents, eventDataNewBlock.ResultFinalizeBlock.Events)
+			require.Equal(t, expectedTxResults, eventDataNewBlock.ResultFinalizeBlock.TxResults)
+			require.Equal(t, postBuildInfo.GetLastBlockAppHash(), eventDataNewBlock.ResultFinalizeBlock.AppHash)
```
joshklop, line 196: Can we assert that the `bfttypes.EventDataNewBlock` struct is equal in one go, rather that testing equality on each of its fields? This way, if new fields are added later, the test will fail (as it should).

I didn't catch this in the first review.

joshklop (approved):

builder/builder_test.go
natebeauregard: Good idea, I added this in a new commit

PR #200, opened 2024-09-10
Add ParentHash to CometBFT header conversion

### LINKED ISSUE #198: Convert monomer header parent hash in ToComet
### Background
Currently, the monomer parent hash is not [converted](https://github.com/polymerdao/monomer/blob/main/monomer.go#L61-L68) whenever converting a monomer header to a comet header. This is important to convert as well since we want it visible in the emitted comet block events.

### Changes
Convert the monomer parent hash as well when converting to a CometBFT header.

AnkushinDaniil:
- Include ParentHash in LastBlockID field of CometBFT header
- Set PartSetHeader Total to 0 and Hash to nil

This change ensures that the parent block information is properly represented when converting from Monomer to CometBFT headers. Including the ParentHash in the LastBlockID allows for correct block linking in
the CometBFT chain.

The PartSetHeader is set with Total: 0 and Hash: nil  because our implementation doesn't use CometBFT's block partitioning feature.

https://github.com/cometbft/cometbft/blob/v0.37.x/spec/core/data_structures.md#header

Fixes #198

## Summary by CodeRabbit

- **New Features**
	- Enhanced blockchain state information by adding a `LastBlockID` field to the header output, providing users with additional context about the last block's identifier.
  
- **Style**
	- Reformatted existing fields in the header output for improved consistency and readability.

### DIFF
--- a/monomer.go
+++ b/monomer.go
@@ -60,10 +60,11 @@ type Header struct {
 
 func (h *Header) ToComet() *bfttypes.Header {
 	return &bfttypes.Header{
-		ChainID: h.ChainID.String(),
-		Height:  int64(h.Height),
-		Time:    time.Unix(int64(h.Time), 0),
-		AppHash: h.StateRoot.Bytes(),
+		ChainID:     h.ChainID.String(),
+		Height:      int64(h.Height),
+		Time:        time.Unix(int64(h.Time), 0),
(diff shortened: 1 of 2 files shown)


AnkushinDaniil:
@natebeauregard

natebeauregard (approved):

PR #201, opened 2024-09-10
Use geth internal GetProof for monomer eth_getProof

natebeauregard:
### Background
Monomer was previously using a lightly modifying version of geth's internal `GetProof` func for the monomer `eth_getProof` endpoint. To minimize duplicated code and the potential for errors, the original geth `GetProof` func was copied to monomer's internal `ethapi` package and all monomer specific logic is now included in a custom monomer ETH API Backend.

### Changes
- Use geth internal GetProof for `eth_getProof`
  - Copies geth `eth_getProof` impl and helper funcs into the `ethapi` package
  - Adds a monomer eth API backend to handle monomer specific logic when calling the internal `eth_getProof` RPC method
  - Removes `get_proof.go` and call into the internal geth API for `eth_getProof` from eth.go
- Rename eth API helpers to include the API suffix

## Summary by CodeRabbit

- **New Features**
	- Introduced `EthAPIBackend` for enhanced Ethereum API interactions, including block header and state retrieval.
	- Added `BlockChainAPI` for blockchain data retrieval and proof generation.
	- New `ProofAPI` type to manage blockchain proofs.

- **Improvements**
	- Renamed types and functions for clarity, transitioning to an API-centric design.
	- Enhanced error handling and modularity in the codebase.

- **Testing Enhancements**
	- Updated test cases to utilize new API constructors, improving consistency and maintainability.


eth/api_backend.go
```diff
@@ -0,0 +1,93 @@
...
+	header := block.Header
+	if header == nil {
+		return nil, fmt.Errorf("header %w", ethereum.NotFound)
+	}
+	return header.ToEth(), nil
+}
+
+func (*EthAPIBackend) HistoricalRPCService() *rpc.Client {
+	return nil
+}
```
joshklop, line 73: Why do we need this?

eth/api_backend.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/polymerdao/monomer"
+	"github.com/polymerdao/monomer/eth/internal/ethapi"
+)
+
+type EthAPIBackend struct {
```
joshklop, line 17: can this be un-exported?

eth/api_backend.go
```diff
@@ -0,0 +1,93 @@
...
+	database   state.Database
+	blockStore DB
+}
+
+func NewEthAPIBackend(db state.Database, blockStore DB) *EthAPIBackend {
```
joshklop, line 22: same, un-exported

joshklop (approved):
Approving pending a few minor fixes, and assuming `internal/ethapi/api.go` continues to match the geth implementation 1:1

eth/api_backend.go
natebeauregard: We'll never hit it because our chain config will always be post bedrock, however it's used [here](https://github.com/ethereum-optimism/op-geth/blob/3653ceb/internal/ethapi/api.go#L713-L724) in the op-geth impl.

natebeauregard:
@joshklop I addressed your code review comments and included them in the first commit before I pushed back up. I ran a diff of our copied op-geth api and the original op-geth api at commit 3653ceb and reordered some of the pre-existing api funcs we had to match the ordering from the op-geth api. I also added an additional commit to update the README link to redirect to the correct op-geth commit.

eth/eth.go
```diff
@@ -88,3 +92,23 @@ func (e *Block) toRPCBlock(block *monomer.Block, fullTx bool) (map[string]any, e
...
+func NewProofProvider(db state.Database, blockStore DB) *ProofProvider {
+	return &ProofProvider{
+		blockchainAPI: ethapi.NewBlockChainAPI(newEthAPIBackend(db, blockStore)),
+	}
+}
+
+// GetProof returns the account and storage values of the specified account including the Merkle-proof.
+// The user can specify either a block number or a block hash to build to proof from.
+func (p *ProofProvider) GetProof(
+	address common.Address,
```
joshklop, line 109: We should accept a `context.Context` here as well (it is permissible with the geth RPC server).

joshklop (approved):

eth/eth.go
natebeauregard: Good catch, I'll add that in the first commit and then merge

PR #208, opened 2024-09-12
Refactor L1 user deposit processing in x/rollup module

### LINKED ISSUE #205: Deposits: verify correct data processing in `x/rollup` `processL1UserDeposits()`
I suspect (85% confidence) the unmarshal of deposit tx data in [processL1UserDepositTxs](https://github.com/polymerdao/monomer/blob/cff062fbca676958cb11167a1c8f2fee6a4eabbf/x/rollup/keeper/deposits.go#L53) is not recovering the data it actually wants.

Specifically:

https://github.com/polymerdao/monomer/blob/cff062fbca676958cb11167a1c8f2fee6a4eabbf/x/rollup/keeper/deposits.go#L59

`tx` is a vanilla Ethereum transaction. We access its `To` and `Value` fields and treat them as if they were the specified parameters of the deposit.
https://github.com/polymerdao/monomer/blob/cff062fbca676958cb11167a1c8f2fee6a4eabbf/x/rollup/keeper/deposits.go#L73-L79

https://github.com/polymerdao/monomer/blob/cff062fbca676958cb11167a1c8f2fee6a4eabbf/x/rollup/keeper/deposits.go#L81

... but they are not. Here, `To` will be the address of the L1OptimismPortal contract. `Value` will be the eth value of the depositing tx, which is distinct from the minting order of the deposit.

Instead, we want the values from the encoded calldata of that tx.

AnkushinDaniil:
Partially fixes #205

## Summary by CodeRabbit

- **New Features**
	- Enhanced configuration capabilities for blockchain applications with the introduction of a new chain configuration function.
	- Improved handling of Ethereum user deposit transactions for enhanced accuracy and compliance with Ethereum standards.
	- Updated minting logic for better representation of mint amounts.
	- Simplified deposit transactions by removing specific deposit amounts, aligning with future changes in transaction handling.

- **Bug Fixes**
	- Enhanced error handling with detailed logging for failures in determining the sender's address, improving clarity in transaction processing issues.

### DIFF
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -314,7 +314,9 @@ func TestBuildRollupTxs(t *testing.T) {
 	require.NotNil(t, depositTxETH.Mint())
 	require.NotNil(t, depositTxETH.To(), "Deposit transaction must have a 'to' address")
 
-	cosmAddr := utils.EvmToCosmosAddress(*depositTxETH.To())
+	from, err := gethtypes.NewLondonSigner(depositTxETH.ChainId()).Sender(depositTxETH)
+	require.NoError(t, err)
+	cosmAddr := utils.EvmToCosmosAddress(from)
 	withdrawalTx := testapp.ToTx(t, &types.MsgInitiateWithdrawal{
 		Sender:   cosmAddr.String(),
 		Target:   common.HexToAddress("0x12345abcde").String(),
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -311,7 +311,7 @@ func TestBuildRollupTxs(t *testing.T) {
 	require.Len(t, ethTxs, 2, "Expected two Ethereum transactions: L1 attribute and deposit tx, got %d", len(ethTxs))
 
 	depositTxETH := ethTxs[1]
-	require.NotNil(t, depositTxETH.Value())
+	require.NotNil(t, depositTxETH.Mint())
 	require.NotNil(t, depositTxETH.To(), "Deposit transaction must have a 'to' address")
 
 	cosmAddr := utils.EvmToCosmosAddress(*depositTxETH.To())
@@ -346,7 +346,7 @@ func TestBuildRollupTxs(t *testing.T) {
 	builtBlock, preBuildInfo, postBuildInfo := buildBlock(t, b, env.app, payload)
 
 	// Test deposit was received
-	checkDepositTxResult(t, env.txStore, depositTxs, fmt.Sprintf("%sETH", depositTxETH.Value().String()), cosmAddr.String())
+	checkDepositTxResult(t, env.txStore, depositTxs, fmt.Sprintf("%sETH", depositTxETH.Mint().String()), cosmAddr.String())
 
 	withdrawalTxResult, err := env.txStore.Get(bfttypes.Tx(withdrawalTx).Hash())
 	require.NoError(t, err)
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -314,7 +314,7 @@ func TestBuildRollupTxs(t *testing.T) {
 	require.NotNil(t, depositTxETH.Mint())
 	require.NotNil(t, depositTxETH.To(), "Deposit transaction must have a 'to' address")
 
-	from, err := gethtypes.NewLondonSigner(depositTxETH.ChainId()).Sender(depositTxETH)
+	from, err := gethtypes.NewCancunSigner(depositTxETH.ChainId()).Sender(depositTxETH)
 	require.NoError(t, err)
 	cosmAddr := utils.EvmToCosmosAddress(from)
 	withdrawalTx := testapp.ToTx(t, &types.MsgInitiateWithdrawal{
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -320,7 +320,7 @@ func TestBuildRollupTxs(t *testing.T) {
 	withdrawalTx := testapp.ToTx(t, &types.MsgInitiateWithdrawal{
 		Sender:   cosmAddr.String(),
 		Target:   common.HexToAddress("0x12345abcde").String(),
-		Value:    math.NewIntFromBigInt(depositTxETH.Value()),
+		Value:    math.NewIntFromBigInt(depositTxETH.Mint()),
 		GasLimit: big.NewInt(100_000).Bytes(),
 		Data:     []byte{},
 	})
--- a/testutils/utils.go
+++ b/testutils/utils.go
@@ -76,9 +76,9 @@ func GenerateEthTxs(t *testing.T) (*gethtypes.Transaction, *gethtypes.Transactio
 	l1InfoTx := gethtypes.NewTx(l1InfoRawTx)
 
 	rng := rand.New(rand.NewSource(1234))
-	inner := testutils.GenerateDeposit(testutils.RandomHash(rng), rng)
-	inner.Mint = big.NewInt(100)
-	depositTx := gethtypes.NewTx(inner)
+	depositRawTx := testutils.GenerateDeposit(testutils.RandomHash(rng), rng)
+	depositRawTx.Mint = big.NewInt(100)
+	depositTx := gethtypes.NewTx(depositRawTx)
 
 	cosmosEthTx := monomer.AdaptNonDepositCosmosTxToEthTx([]byte{1})
 	return l1InfoTx, depositTx, cosmosEthTx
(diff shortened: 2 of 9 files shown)


AnkushinDaniil:
@natebeauregard

testutils/utils.go
```diff
@@ -75,7 +75,7 @@ func GenerateEthTxs(t *testing.T) (*gethtypes.Transaction, *gethtypes.Transactio
 	require.NoError(t, err)
 	l1InfoTx := gethtypes.NewTx(l1InfoRawTx)
 
-	rng := rand.New(rand.NewSource(1234))
+	rng := rand.New(rand.NewSource(12345))
```
natebeauregard, line 78: Instead of modifying the seed could you just set the `Mint` field directly in the test helper? We'll also want to test the case where `Mint` is nil
You should be able to modify it [here](https://github.com/polymerdao/monomer/blob/ddcdad2bd95fd01645f7225f8ab770569b56741a/testutils/utils.go#L79) before calling `gethtypes.NewTx`

x/rollup/keeper/deposits.go
```diff
@@ -70,15 +70,23 @@ func (k *Keeper) processL1UserDepositTxs(ctx sdk.Context, txs [][]byte) (sdk.Eve
...
+		if tx.To() == nil {
 			ctx.Logger().Error("Contract creation txs are not supported", "index", i)
 			return nil, types.WrapError(types.ErrInvalidL1Txs, "Contract creation txs are not supported, index:%d", i)
 		}
-		cosmAddr := utils.EvmToCosmosAddress(*to)
-		mintAmount := sdkmath.NewIntFromBigInt(tx.Value())
+
+		// Get the sender's address from the transaction
+		// We use NewLondonSigner because deposit transactions follow EIP-2718 and EIP-2930
+		to, err := ethtypes.NewLondonSigner(tx.ChainId()).Sender(&tx)
```
natebeauregard, line 82: Can you call this variable `from` or `sender` instead to differentiate it from the deposit tx `To` field?

builder/builder_test.go
```diff
@@ -311,15 +311,17 @@ func TestBuildRollupTxs(t *testing.T) {
 	require.Len(t, ethTxs, 2, "Expected two Ethereum transactions: L1 attribute and deposit tx, got %d", len(ethTxs))
 
 	depositTxETH := ethTxs[1]
-	require.NotNil(t, depositTxETH.Value())
+	require.NotNil(t, depositTxETH.Mint())
 	require.NotNil(t, depositTxETH.To(), "Deposit transaction must have a 'to' address")
 
-	cosmAddr := utils.EvmToCosmosAddress(*depositTxETH.To())
+	to, err := gethtypes.NewLondonSigner(depositTxETH.ChainId()).Sender(depositTxETH)
```
natebeauregard, line 317: can you update the variable name here as well from `to` to `from` or `sender`?

builder/builder_test.go
```diff
@@ -311,15 +311,17 @@ func TestBuildRollupTxs(t *testing.T) {
...
+	require.NoError(t, err)
+	cosmAddr := utils.EvmToCosmosAddress(to)
```
natebeauregard, line 319: nit: might be good to just inline `cosmAddr` and remove the variable since it's only used once

x/rollup/tests/integration/rollup_test.go
```diff
@@ -40,14 +41,16 @@ func TestRollup(t *testing.T) {
 	l1AttributesTxBz := monomertestutils.TxToBytes(t, l1AttributesTx)
 	depositTxBz := monomertestutils.TxToBytes(t, depositTx)
 
-	depositAmount := depositTx.Value()
-	var userAddr sdk.AccAddress = depositTx.To().Bytes()
+	depositAmount := depositTx.Mint()
+	to, err := gethtypes.NewLondonSigner(depositTx.ChainId()).Sender(depositTx)
```
natebeauregard, line 45: can you update the variable name here as well from `to` to `from` or `sender`?

x/rollup/tests/integration/rollup_test.go
```diff
@@ -40,14 +41,16 @@ func TestRollup(t *testing.T) {
...
+	require.NoError(t, err)
+	var userAddr sdk.AccAddress = to.Bytes()
```
natebeauregard, line 47: nit: we could use `utils.EvmToCosmosAddress(sender.Bytes())` here and just inline it

x/rollup/keeper/deposits.go
```diff
@@ -70,15 +70,23 @@ func (k *Keeper) processL1UserDepositTxs(ctx sdk.Context, txs [][]byte) (sdk.Eve
...
-		if to == nil {
+		if tx.To() == nil {
 			ctx.Logger().Error("Contract creation txs are not supported", "index", i)
 			return nil, types.WrapError(types.ErrInvalidL1Txs, "Contract creation txs are not supported, index:%d", i)
 		}
-		cosmAddr := utils.EvmToCosmosAddress(*to)
-		mintAmount := sdkmath.NewIntFromBigInt(tx.Value())
+
+		// Get the sender's address from the transaction
+		// We use NewLondonSigner because deposit transactions follow EIP-2718 and EIP-2930
```
natebeauregard, line 81: I'm not sure if it makes a difference but out of curiosity why did you pick the London signer? Should we use Cancun instead?

cc @joshklop

testutils/utils.go
natebeauregard: Alternatively, should we just use a random seed for every test so we can check that our tests pass for all deposit tx variations?

cc @joshklop

x/rollup/tests/integration/rollup_test.go
natebeauregard: oh lol I think I wrote this initial code, imo we should still update it though if you don't mind

builder/builder_test.go
AnkushinDaniil: We use it twice

x/rollup/tests/integration/rollup_test.go
AnkushinDaniil: We use `userAddr` more than once

x/rollup/keeper/deposits.go
AnkushinDaniil: We can try to use `MakeSigner`, but we need `*params.ChainConfig` for that.

testutils/utils.go
AnkushinDaniil: Do you want me to implement a test for the case when `Mint` is nil in this or another PR?

x/rollup/keeper/deposits.go
joshklop: Perhaps we should move this to a global variable in the `monomer` package, until we allow more dynamic configuration on the OP Stack side.

https://github.com/polymerdao/monomer/blob/609bc9995b12777ed2d76c789bf422211bf4d586/eth/internal/ethapi/helpers.go#L37-L56

testutils/utils.go
natebeauregard: Hmm actually for now let's just leave `Mint` as always set and then we should open up a new issue to ensure that we have some fuzz testing for all possible deposit tx types (with and without a mint field, with and without a value field, etc.). It probably doesn't make sense to fuzz test all of our unit/integration tests but it's definitely good to keep in mind

x/rollup/tests/integration/rollup_test.go
natebeauregard: Ok sounds good. Could we still update it to use `utils.EvmToCosmosAddress(from.Bytes())`?

natebeauregard (commented):
Looking good, left a few small nits. I'll wait for @joshklop's chain config question to be addressed before approving

x/rollup/keeper/deposits.go
joshklop: Ah, we probably need `NewChainConfig` to be a function in the `monomer` package since we need to create a different chain config depending on the chain id.

x/rollup/keeper/deposits.go
```diff
@@ -78,7 +80,11 @@ func (k *Keeper) processL1UserDepositTxs(ctx sdk.Context, txs [][]byte) (sdk.Eve
 		}
 
 		// Get the sender's address from the transaction
-		from, err := ethtypes.NewCancunSigner(tx.ChainId()).Sender(&tx)
+		from, err := ethtypes.MakeSigner(
+			monomer.NewChainConfig(tx.ChainId()),
+			big.NewInt(ctx.BlockHeight()),
+			uint64(ctx.BlockTime().Unix()),
```
natebeauregard, line 86: Will `ctx` will store the L2 block height and block time? Should we be using the L1 block height/block time instead here?

natebeauregard: I think that you should be able to get the L1 block height/time from the L1 block info stored in the `x/rollup` store

natebeauregard: Yeah I'll reach out over slack

natebeauregard (approved):

PR #211, opened 2024-09-13
Add-guard-against-deposit-transaction

AnkushinDaniil:
## Summary by CodeRabbit

- **New Features**
	- Enhanced transaction validation by preventing deposit transactions from being enqueued in the mempool.
  
- **Bug Fixes**
	- Improved error handling for deposit transactions within the mempool.

- **Tests**
	- Added a new test case to ensure the mempool correctly handles deposit transactions, verifying that they are not accepted.

### DIFF
--- a/mempool/mempool_test.go
+++ b/mempool/mempool_test.go
@@ -31,8 +31,7 @@ func TestMempool(t *testing.T) {
 
 		cosmosTxs, err := monomer.AdaptPayloadTxsToCosmosTxs([]hexutil.Bytes{depositTxBytes}, nil, "")
 		require.NoError(t, err)
-
-		require.Error(t, pool.Enqueue(cosmosTxs[0]))
+		require.ErrorIs(t, pool.Enqueue(cosmosTxs[0]), mempool.ErrDepositTxInPool)
 	})
 
 	// enqueue multiple to empty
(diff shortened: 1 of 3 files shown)


AnkushinDaniil:
@joshklop

joshklop (approved):

mempool/mempool_test.go
```diff
@@ -22,6 +24,17 @@ func TestMempool(t *testing.T) {
...
 
+	t.Run("deposit transaction", func(t *testing.T) {
+		_, depositTx, _ := testutils.GenerateEthTxs(t)
+		depositTxBytes, err := depositTx.MarshalBinary()
+		require.NoError(t, err)
+
+		cosmosTxs, err := monomer.AdaptPayloadTxsToCosmosTxs([]hexutil.Bytes{depositTxBytes}, nil, "")
+		require.NoError(t, err)
+
+		require.Error(t, pool.Enqueue(cosmosTxs[0]))
```
joshklop, line 35: Can we use `ErrorContains` here, and test for the exact error string? It's generally not a great approach, but I think it's warranted here since we want to make sure we're hitting the correct error :)

joshklop: ErrorIs will require us to export the error, which we don't want to do.

AnkushinDaniil: [Change ErrorIs to ErrorContains](https://github.com/polymerdao/monomer/pull/211/commits/55be3960045327498979d4a7b634a99d53c35dc0)

PR #213, opened 2024-09-15
Add monogen tool

joshklop:
## Summary by CodeRabbit

- **New Features**
	- Introduced a command-line tool `monogen` for scaffolding Cosmos SDK projects.
	- Added automation script `monogen.sh` for initializing Cosmos SDK applications.
  
- **Bug Fixes**
	- Updated dependencies to enhance compatibility and security.

- **Chores**
	- Enhanced `.gitignore` to manage ignored files more effectively.
	- Updated build process in the Makefile for better binary management.


joshklop:
To test yourself, run 
```bash
make monogen \
    && cd .. \
    && ./monomer/bin/monogen \
    && cd testapp \
    && ../monomer/monogen/monogen.sh \
    && ./testappd monomer start --minimum-gas-prices 0.01stake
```

Two follow-ups are planned:
1. Docs updates
2. Generating a realistic testapp for e2e tests.

The idea is that we can instruct users to run
```bash
go run github.com/polymerdao/monomer/cmd/monogen@latest
```
and automatically paste `monogen.sh` in the docs so we never have to modify the tutorial. This has the added benefit of running the tutorial in tests, so it will always be up-to-date.

joshklop:
I'll fix the lints tomorrow. This is ready for review otherwise.

monogen/cmd/main.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	skipGit       bool
+	appDirPath    string
+	goModulePath  string
+	addressPrefix string
+)
+
+func main() {
+	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGTERM)
```
natebeauregard, line 32: Should `syscall.SIGINT` also be included in the `NotifyContext` params in case the user interrupts the tool with Ctrl-C?

monogen/monogen.go
```diff
@@ -0,0 +1,263 @@
...
+	// TODO: don't do anything if directory already exists. I feel like ignite should handle this somewhere.
+
+	igniteRootDir, err := config.DirPath()
+	if err != nil {
+		return fmt.Errorf("get ignite config directory: %v", err)
+	}
+	// https://github.com/ignite/cli/blob/2a968e8684cae0a1d79ccb11f0db067a4605173e/ignite/cmd/cmd.go#L33-L34
+	cacheDBPath := filepath.Join(igniteRootDir, "ignite_cache.db")
+	// Note that we need to change this version every time we bump ignite's version in Monomer's go.mod.
+	cacheStorage, err := cache.NewStorage(cacheDBPath, cache.WithVersion("v28.5.1"))
```
natebeauregard, line 28: Would it make sense to just read the ignite version directly from `go.mod` to avoid needing to change this manually if the ignite version is ever updated?

monogen/monogen.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/ignite/cli/v28/ignite/pkg/cache"
+	"github.com/ignite/cli/v28/ignite/pkg/gocmd"
+	"github.com/ignite/cli/v28/ignite/pkg/placeholder"
+	"github.com/ignite/cli/v28/ignite/pkg/xast"
+	"github.com/ignite/cli/v28/ignite/services/scaffolder"
+	"github.com/ignite/cli/v28/ignite/templates/module"
+)
+
+func Generate(ctx context.Context, appDirPath, goModulePath, addressPrefix string, skipGit bool) error {
+	// TODO: don't do anything if directory already exists. I feel like ignite should handle this somewhere.
```
natebeauregard, line 19: Is addressing this TODO going to be one of the follow up PRs for this script?

monogen/monogen.go
```diff
@@ -0,0 +1,263 @@
...
+		return fmt.Errorf("find: %v", err)
+	}
+	content := replacer.Replace(goMod.String(), "replace (", `replace (
+	cosmossdk.io/core => cosmossdk.io/core v0.11.1
+	github.com/btcsuite/btcd/btcec/v2 v2.3.4 => github.com/btcsuite/btcd/btcec/v2 v2.3.2
+	github.com/crate-crypto/go-ipa => github.com/crate-crypto/go-ipa v0.0.0-20231205143816-408dbffb2041
+	github.com/crate-crypto/go-kzg-4844 v1.0.0 => github.com/crate-crypto/go-kzg-4844 v0.7.0
+	github.com/ethereum/go-ethereum => github.com/joshklop/op-geth v0.0.0-20240515205036-e3b990384a74
+	github.com/libp2p/go-libp2p => github.com/joshklop/go-libp2p v0.0.0-20240814165419-c6b91fa9f263
+`)
```
natebeauregard, line 258: optional - could we read the replace directives directly from `go.mod` as well so we don't need to update the script every time they're updated?

monogen/monogen_test.go
```diff
@@ -0,0 +1,69 @@
...
+	defer cancel()
+	cmd := initCommand(exec.CommandContext(ctx, filepath.Join(appDirPath, appName+"d"), "monomer", "start", "--minimum-gas-prices", "0.0001stake"))
+	require.NoError(t, cmd.Start())
+	defer func() {
+		cancel()
+		// TODO: handle this error.
+		// It is kind of harmless, just `signal: killed` or `signal: terminated`,
+		// but it would be nice to be able to say `require.NoError` here.
+		_ = cmd.Wait()
+	}()
```
natebeauregard, line 49: Why is this `defer func` necessary? Isn't the context already properly canceled with the `defer cancel()` call above?

Makefile
```diff
@@ -2,13 +2,18 @@ GOBIN ?= $$(go env GOPATH)/bin
...
 
 E2E_ARTIFACTS_PATH ?= e2e/artifacts
 E2E_STATE_SETUP_PATH ?= e2e/optimism/.devnet
 E2E_CONFIG_SETUP_PATH ?= e2e/optimism/packages/contracts-bedrock/deploy-config/devnetL1.json
 FOUNDRY_ARTIFACTS_PATH ?= bindings/artifacts
 FOUNDRY_CACHE_PATH ?= bindings/cache
 
+.PHONY: monogen
+monogen:
+	go build -o $(BIN)/monogen ./monogen/cmd
```
natebeauregard, line 15: The ignite tool generates a bunch of comments used by starport scaffolding (e.g. `// this line is used by starport scaffolding # stargate/app/maccPerms`).

There are a few cases in the generated `testapp/app_config` file where the comment is overwritten by the monogen script. I don't know too much about how startport scaffolding works but do you know of any potential errors that could arise from overwriting that comment?

natebeauregard (approved):
LGTM! Left a few minor nits and questions but the impl looks solid

monogen/monogen_test.go
joshklop: `defer`s are executed in reverse order.

`Wait` will only finish if `cancel` is called first.

This means that `cancel` is called twice on the happy path.

The first `defer cancel()` just ensures `cancel` is called on even the strangest unhappy paths.

monogen/monogen.go
joshklop: We would need the path to our own `go.mod`, which may not even be present on a user's machine.

However, we might be able to embed the `go.mod` in the `monogen.go` file and read it directly...

This is actually a really interesting idea. Will try.

monogen/cmd/main.go
joshklop: Yes, thanks.

monogen/monogen.go
joshklop: The replace directives for apps are different than the replace directives for Monomer.

Makefile
joshklop: Ohh overwriting the comment shouldn't happen. Possibly an oversight on my end--we're supposed to add the comment back. Will double check on my end.

monogen/monogen.go
joshklop: I forgot we can't `go:embed` files from parent directories, so we'd need to either make monogen a separate module or copy the `go.mod` file to the `monogen` directory. Neither option is super appealing to me.

I decided to use the `version.Version` variable from ignite, which is set to `development` when accessed programatically. When the ignite binary is built, this variable is set to the release tag, so we won't be able to use the same cache, but at least we shouldn't corrupt anything.

natebeauregard (approved):

PR #214, opened 2024-09-16
Update gomock

AnkushinDaniil:
This pull request moves the generation command from the makefile to `expected_keepers.go` and updates gomock to the maintained fork.

This may be useful in the future as the use of mocks expands.


AnkushinDaniil:
@natebeauregard

natebeauregard (commented):
@AnkushinDaniil Thanks for making this PR. I talked with @joshklop and we agreed that this change adds some additional indirection that may make it harder to determine how our mocks are being generated and also makes our mock generation differ from the traditional cosmos sdk and most major cosmos projects. That being said, I don't think these changes will be a valuable addition to the repo.

However, I like that you discovered that the golang mock repo is archived and like the switch to using the maintained uber fork. Could you repurpose this PR to just update the mockgen dependency? We'll want to lock it to a specific version as well instead of using the latest, could we use `v0.4.0`?

natebeauregard (approved):

PR #215, opened 2024-09-16
Make config & genesis state data available to op-e2e.config init() function

### LINKED ISSUE #207: [e2e] `allocs-l1.json does not exist, skipping genesis generation` on e2e test
The following log line appeared on my machine during e2e testing, which I believe is new:

```
file /home/colin/nm/polymerdao/monomer/.devnet/allocs-l1.json does not exist, skipping genesis generation
```

It is also happening on CI:

https://github.com/polymerdao/monomer/actions/runs/10822155982/job/30025536688#step:6:6

The error comes from `optimism.op-e2e.config.init()` - the init function of the `config` package, and it causes an [early exit](https://github.com/ethereum-optimism/optimism/blob/24a8d3e06e61c7a8938dfb7a591345a437036381/op-e2e/config/init.go#L107) from this init. The `init()` seems to make an assumption about the running directory of whatever process invokes it, but this assumption is not true in our e2e test.

It's not clear to me if we ever use the `op-e2e.config` package, but something must, in order to invoke the `init()` function.

NiloCK:
PR closes #207.

PR is somewhat clumsy. It:
- modifies `make.setup-e2e` to
  - make a copy of `./.devnet` in the root directory
  - make a copy of the `devnetL1.json` config file in `./packages/contracts-bedrock/deploy-config/`
- gitignores both sets of files

Both copied sets are referenced, at that location, by the `op-e2e.config.init()` function.

## Summary by CodeRabbit

- **Chores**
	- Updated `.gitignore` to exclude unnecessary development and deployment files.
	- Enhanced the `setup-e2e` process in the Makefile for improved end-to-end testing setup, ensuring necessary configuration files are readily available.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,7 @@ e2e/artifacts
 # Ignore the forge artifacts and cache.
 bindings/artifacts
 bindings/cache
+
+# Ignore the op stack genesis outputs from make setup-e2e
+.devnet
+packages/contracts-bedrock/deploy-config/
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,6 @@ setup-e2e:
 	$(MAKE) -C e2e/optimism install-geth && \
 	$(MAKE) -C e2e/optimism cannon-prestate && \
 	$(MAKE) -C e2e/optimism devnet-allocs && \
-	# Make genesis state and config available to op-e2e/config/init.go. See issue 207. && \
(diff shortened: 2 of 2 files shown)


joshklop:
It turns out we can suppress this error and remove ~70 lines from our own e2e tests by building on top of the op-e2e initialization. Let's continue the discussion at #216

Issue #210, opened 2024-09-12
Investigate depositing ERC-20s through x/rollup


natebeauregard:
We'll want to eventually support depositing ERC-20 tokens to monomer based L2s, we should investigate how we could modify monomer and the `x/rollup` module deposit flow to support this.

PR #223, opened 2024-09-17
Support ERC-20 deposits in x/rollup

natebeauregard:
Adds support for ERC-20 deposits by parsing the deposit tx data. ERC-20 tokens are minted on L2 in the `erc20/{L1 ERC-20 address}` format.

Safeguards are in place to ensure that the deposit tx originated from the L1CrossDomainMessenger so that users can't forge tx data to mint unlimited ERC-20 tokens.

`x/rollup` integration tests and e2e tests with ERC-20 deposits were added to verify the deposit was successful.

This is an initial increment to get a functional implementation for ERC-20 deposits. Future work to either add a genesis state to `x/rollup` to contain the `L1CrossDomainMessenger` address or to extract the ERC-20 deposit logic to a new cosmos module will be done in a future increment.

Closes https://github.com/polymerdao/monomer/issues/210

## Summary by CodeRabbit

- **New Features**
	- Introduced structures for managing Ethereum ERC20 token deposits and relay messaging.
	- Enhanced testing suite with specific functions for handling ETH and ERC20 token deposits.
	- Added support for minting ERC20 tokens and processing cross-domain messages.
	- Improved integration with Ethereum Layer 1 components.
	- Expanded event handling capabilities to include ERC20 token operations.

- **Bug Fixes**
	- Improved error handling and logging in transaction processing.

### DIFF
--- a/e2e/stack_test.go
+++ b/e2e/stack_test.go
@@ -265,7 +265,7 @@ func ethRollupFlow(t *testing.T, stack *e2e.StackConfig) {
 
 	// send user Deposit Tx
 	depositAmount := big.NewInt(params.Ether)
-	depositTx, err := stack.L1Portal.DepositTransaction(
+	depositTx, err := stack.OptimismPortal.DepositTransaction(
 		createL1TransactOpts(t, stack, userPrivKey, l1signer, l1GasLimit, depositAmount),
 		userAddress,
 		big.NewInt(0),
@@ -287,7 +287,7 @@ func ethRollupFlow(t *testing.T, stack *e2e.StackConfig) {
 	require.NotNil(t, receipt, "deposit tx receipt")
 	require.Equal(t, types.ReceiptStatusSuccessful, receipt.Status, "deposit tx reverted")
 
-	depositLogs, err := stack.L1Portal.FilterTransactionDeposited(
+	depositLogs, err := stack.OptimismPortal.FilterTransactionDeposited(
 		&bind.FilterOpts{
 			Start:   0,
 			End:     nil,
@@ -357,7 +357,7 @@ func ethRollupFlow(t *testing.T, stack *e2e.StackConfig) {
 	require.NoError(t, err)
 
 	// send a withdrawal proving tx to prove the withdrawal on L1
-	proveWithdrawalTx, err := stack.L1Portal.ProveWithdrawalTransaction(
+	proveWithdrawalTx, err := stack.OptimismPortal.ProveWithdrawalTransaction(
 		createL1TransactOpts(t, stack, userPrivKey, l1signer, l1GasLimit, nil),
 		withdrawalTx.WithdrawalTransaction(),
 		provenWithdrawalParams.L2OutputIndex,
@@ -377,7 +377,7 @@ func ethRollupFlow(t *testing.T, stack *e2e.StackConfig) {
 
 	withdrawalTxHash, err := withdrawalTx.Hash()
 	require.NoError(t, err)
-	proveWithdrawalLogs, err := stack.L1Portal.FilterWithdrawalProven(
+	proveWithdrawalLogs, err := stack.OptimismPortal.FilterWithdrawalProven(
 		&bind.FilterOpts{
 			Start:   0,
 			End:     nil,
@@ -401,7 +401,7 @@ func ethRollupFlow(t *testing.T, stack *e2e.StackConfig) {
 	require.NoError(t, err)
 
 	// send a withdrawal finalizing tx to finalize the withdrawal on L1
-	finalizeWithdrawalTx, err := stack.L1Portal.FinalizeWithdrawalTransaction(
+	finalizeWithdrawalTx, err := stack.OptimismPortal.FinalizeWithdrawalTransaction(
 		createL1TransactOpts(t, stack, userPrivKey, l1signer, l1GasLimit, nil),
 		withdrawalTx.WithdrawalTransaction(),
 	)
@@ -416,7 +416,7 @@ func ethRollupFlow(t *testing.T, stack *e2e.StackConfig) {
 	require.NotNil(t, receipt, "finalize withdrawal tx receipt")
 	require.Equal(t, types.ReceiptStatusSuccessful, receipt.Status, "finalize withdrawal tx failed")
 
-	finalizeWithdrawalLogs, err := stack.L1Portal.FilterWithdrawalFinalized(
+	finalizeWithdrawalLogs, err := stack.OptimismPortal.FilterWithdrawalFinalized(
 		&bind.FilterOpts{
 			Start:   0,
 			End:     nil,
@@ -515,7 +515,7 @@ func erc20RollupFlow(t *testing.T, stack *e2e.StackConfig) {
 	require.NoError(t, err)
 
 	// check that the deposit tx went through the OptimismPortal successfully
-	_, err = receipts.FindLog(depositReceipt.Logs, stack.L1Portal.ParseTransactionDeposited)
+	_, err = receipts.FindLog(depositReceipt.Logs, stack.OptimismPortal.ParseTransactionDeposited)
 	require.NoError(t, err, "should emit deposit event")
 
 	// assert the user's bridged WETH is no longer on L1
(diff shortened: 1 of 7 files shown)


x/rollup/keeper/deposits.go
```diff
@@ -109,3 +184,60 @@ func (k *Keeper) mintETH(ctx sdk.Context, addr sdk.AccAddress, amount sdkmath.In
...
+		sdk.NewAttribute(types.AttributeKeyValue, hexutil.Encode(amount.BigInt().Bytes())),
+	)
+
+	return &mintEvent, nil
+}
+
+// TODO: should this belong in the utils package?
+// unpackInputsIntoInterface unpacks the input data of a function call into an interface. This function behaves
+// similarly to the geth abi UnpackIntoInterface function but unpacks method inputs instead of outputs.
+func unpackInputsIntoInterface(contractABI *abi.ABI, methodName string, inputData []byte, outputInterface interface{}) error {
```
natebeauregard, line 218: TODO: Should we just swallow all errors here? Or should we only swallow errors for unrecognized L1StandardBridge messages?

x/rollup/keeper/deposits.go
```diff
@@ -85,19 +94,85 @@ func (k *Keeper) processL1UserDepositTxs(ctx sdk.Context, txs [][]byte) (sdk.Eve
...
+		}
 	}
 
 	return mintEvents, nil
 }
 
+// parseAndExecuteCrossDomainMessage parses the tx data of a cross domain message and applies state transitions for recognized messages.
+// Currently, only finalizeBridgeERC20 messages from the L1StandardBridge are recognized for minting ERC-20 tokens on the Cosmos chain.
+// If a message is not recognized, it returns nil and does not error.
+func (k *Keeper) parseAndExecuteCrossDomainMessage(ctx sdk.Context, txData []byte) (*sdk.Event, error) { //nolint:gocritic // hugeParam
```
natebeauregard, line 129: TODO: Should we just swallow all errors here? Or should we only swallow errors for unrecognized L1StandardBridge messages?

natebeauregard:
Force-pushed to rebase on main to get the `docusaurus` CI job passing

x/rollup/keeper/deposits.go
natebeauregard, line 218: Any thoughts from reviewers on if this helper func would be better in the utils package?

natebeauregard:
Force-pushed to rebase on main and resolve merge conflicts

x/rollup/keeper/deposits.go
```diff
@@ -21,6 +29,7 @@ import (
 // Persisted data conforms to optimism specs on L1 attributes:
 // https://github.com/ethereum-optimism/optimism/blob/develop/specs/deposits.md#l1-attributes-predeployed-contract
 func (k *Keeper) setL1BlockInfo(ctx sdk.Context, info derive.L1BlockInfo) error { //nolint:gocritic
+	// TODO: use binary marshaling instead of json
 	infoBytes, err := json.Marshal(info)
```
natebeauregard, line 33: Question to reviewers - Do you think it's worth bringing up an issue for this?

e2e/stack_test.go
```diff
@@ -441,6 +447,97 @@ func rollupFlow(t *testing.T, stack *e2e.StackConfig) {
 	t.Log("Monomer can initiate withdrawals on L2 and can generate proofs for verifying the withdrawal on L1")
 }
 
+func erc20RollupFlow(t *testing.T, stack *e2e.StackConfig) {
+	// TODO: debug race condition with ethRollupFlow/erc20RollupFlow
+	time.Sleep(2 * time.Second)
```
natebeauregard, line 452: question to reviewers - I haven't had any luck in fixing the race condition without this `time.Sleep`. I believe that the race condition occurrs because the `ethRollupFlow/erc20RollupFlow` L1 tx nonces get out of order. To avoid focusing too much on this one particular issue, wdyt about just leaving in this `time.Sleep` for a staggered start to avoid the race condition?

joshklop: I think we should go ahead and solve this. Having functional e2e tests will be the basis of our future work to make sure we don't break anything.

The nonce issue makes sense. To confirm, we can try using a different user than stack.User. Let's just put the whole Secrets struct on the StackConfig and use Secrets.Alice and Secrets.Bob for different tests.

x/rollup/keeper/deposits.go
joshklop: Why is it an issue in the first place? If there is a benefit for binary (size?), let's add a good first issue.

natebeauregard: Yeah mainly just size and performance, we could use protobufs here instead. Sounds good, I'll raise an issue for this

e2e/stack_test.go
natebeauregard: Good point, I feel like that would be a quick fix that would resolve the race condition. I'll look into this

x/rollup/keeper/deposits.go
natebeauregard: https://github.com/polymerdao/monomer/issues/233

x/rollup/keeper/deposits.go
```diff
@@ -126,3 +195,60 @@ func (k *Keeper) mintETH(ctx sdk.Context, addr sdk.AccAddress, amount sdkmath.In
...
+	amount sdkmath.Int,
+) (*sdk.Event, error) {
+	// use the "erc20/{l1erc20addr}" format for the coin denom
+	coin := sdk.NewCoin("erc20/"+erc20addr[2:], amount)
+	if err := k.bankkeeper.MintCoins(ctx, types.ModuleName, sdk.NewCoins(coin)); err != nil {
+		return nil, fmt.Errorf("failed to mint ERC-20 deposit coins to the rollup module: %v", err)
+	}
+	if err := k.bankkeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, userAddr, sdk.NewCoins(coin)); err != nil {
+		return nil, fmt.Errorf("failed to send ERC-20 deposit coins from rollup module to user account %v: %v", userAddr, err)
+	}
```
joshklop, line 212: Can we check what the standard Cosmos way to handle erc20s is? And document it here

joshklop (approved):
No major comments. There are definitely some follow-up items, but it's better to get this on `main` and iterate.

Fantastic work.

x/rollup/keeper/deposits.go
joshklop: Good point. Let's throw on all errors and add a TODO about needing to follow the spec in all cases.

x/rollup/keeper/deposits.go
joshklop: I think there is a helper for this in geth somewhere.

x/rollup/keeper/deposits.go
natebeauregard: https://github.com/polymerdao/monomer/issues/234

x/rollup/keeper/deposits.go
natebeauregard: I could have sworn there was one as well, however I wasn't able to find it after looking for way too long. There is an `UnpackIntoInterface` method that unpacks method outputs instead of inputs and there is a `method.Inputs.UnpackIntoMap`, however that still requires conversion to an interface.

Also, I'm going to move this to the generic utils package since it's not deposit specific, lmk if you're able to find the geth helper or if you're opposed to the move.

x/rollup/keeper/deposits.go
natebeauregard: The only cosmos precedent I was able to find for erc20s is the [x/erc20](https://docs.evmos.org/protocol/modules/erc20#concepts) module from evmos. However, their token pairs need to be pre-registered with a EVM <-> Cosmos token mapping.

Our use case as an L2 makes this a novel problem for us if we want permissionless ERC-20 deposits without any whitelists or denom mapping. I believe that using the `erc20/{l1 address}` format for the cosmos token denom would be a functional and future-proof solution for us.

Please lmk if I misunderstood your question

natebeauregard:
Force-pushed to format commits for merge

x/rollup/keeper/deposits.go
joshklop: Nice, let's go with this.

x/rollup/keeper/deposits.go
joshklop: Ah disregard then.

I'd prefer to keep it private in the keeper package until it's actually needed elsewhere. But not a huge deal either way.

natebeauregard: sgtm, I'll move it back to `deposits.go` and will push back up

PR #249, opened 2024-10-10
Enable transfers during L1 deposits and add L1 wallet integration to the devnet

natebeauregard:
Adds an L1 wallet integration to interact with the devnet Ethereum network and deposit ETH on L2 networks.

Since Metamask (L1) and Keplr (L2) wallets have different private keys, the `x/rollup` deposit flow had to be expanded to supports [transfers on deposits](https://specs.optimism.io/protocol/deposits.html#the-deposited-transaction-type) as well. The amount of transferred ETH is determined by the `Value` field in the deposit tx and is sent to the `To` address in the deposit tx.

Any extra minted ETH, determined by the `Mint` field minus the amount of transferred ETH (`Value`), will be sent to the sender address of the deposit tx.

## Release Notes

- **New Features**
	- Introduced a wallet integration interface with sections for L1 and L2 wallet functionalities, including deposit options for ETH.
	- Added bech32 encoding/decoding support for handling addresses in the browser.
	- Integrated MetaMask for Ethereum deposits, enhancing user experience with transaction management.

- **Bug Fixes**
	- Refined transaction handling in tests to ensure accurate address and amount validation for Ethereum transactions.

- **Documentation**
	- Updated HTML titles and structure to reflect the new wallet integration focus.

- **Chores**
	- Removed outdated configurations related to L1 chain integration, streamlining the focus on L2.


opdevnet/l1.go
```diff
@@ -58,6 +58,8 @@ func (cfg *L1Config) Run(ctx context.Context, env *environment.Env, logger log.L
 		func(_ *ethconfig.Config, nodeCfg *node.Config) error {
 			nodeCfg.WSHost = cfg.URL.Hostname()
 			nodeCfg.WSPort = int(cfg.URL.PortU16())
+			// TODO: make L1 http port configurable
+			nodeCfg.HTTPPort = 39473
```
natebeauregard, line 62: I opted to include this change in a follow-up PR for now, lmk if you think I should update it in this PR instead

x/rollup/types/events.go
```diff
@@ -3,11 +3,13 @@ package types
...
 	AttributeKeyL2WithdrawalTx    = "l2_withdrawal_tx"
-	AttributeKeyFromCosmosAddress = "from_cosmos_address"
-	AttributeKeyToCosmosAddress   = "to_cosmos_address"
+	AttributeKeyFromCosmosAddress = "from_address"
+	AttributeKeyToCosmosAddress   = "to_address"
+	AttributeKeyMintCosmosAddress = "mint_address"
 	AttributeKeySender            = "sender"
 	AttributeKeyL1Target          = "target"
 	AttributeKeyValue             = "value"
+	AttributeKeyMint              = "mint"
```
natebeauregard, line 12: I don't love these event attribute names and feel like they may be confusing. I'll think about the best way to format our mint events and may update these

x/rollup/keeper/deposits.go
```diff
@@ -121,13 +121,15 @@ func (k *Keeper) processL1UserDepositTxs(
 			ctx.Logger().Error("Failed to get sender address", "evmAddress", from, "err", err)
 			return nil, types.WrapError(types.ErrInvalidL1Txs, "failed to get sender address: %v", err)
 		}
-		cosmAddr := utils.EvmToCosmosAddress(from)
+		mintAddr := utils.EvmToCosmosAddress(from)
```
natebeauregard, line 124: question to reviewers - do you think `senderAddr` or `fromAddr` is a better name here than `mintAddr`?

PR #267, opened 2024-10-23
Expand e2e tests to cover the true end-to-end flow

joshklop:
## Release Notes

- **New Features**
  - Introduced a new method for retrieving the chain ID in the MonomerClient.
  - Added functionality for structured end-to-end testing with a new testing framework.

- **Improvements**
  - Streamlined the end-to-end testing setup by removing unnecessary complexity.
  - Updated the README to simplify development prerequisites.
  - Refined handling of L1 allocations and Monomer node initialization.

- **Bug Fixes**
  - Enhanced error handling in various components to improve reliability.

- **Chores**
  - Removed deprecated files and references to improve code cleanliness and maintainability.


joshklop:
We might need to commit a `go.work` file to make the e2e tests pass in the CI. Will investigate later; the tests pass locally for me.

cmd/monogen/testapp/setup-helper.sh
```diff
@@ -6,8 +6,11 @@ set -e
 # Build the app.
 # Because we transitively depend on github.com/fjl/memsize, we need to disable checklinkname in go1.23.0 and higher.
 goVersion=$(go env GOVERSION)
-[[ $goVersion > go1.23.0 || $goVersion == go1.23.0 ]] && ldflags="-ldflags=-checklinkname=0"
-go build $ldflags -o testappd ./cmd/testappd
+if [[ $goVersion > go1.23.0 || $goVersion == go1.23.0 ]]; then
+    GOFLAGS="$GOFLAGS -ldflags=-checklinkname=0"
+fi
+export GOFLAGS
```
natebeauregard, line 12: Will this append `-ldflags=-checklinkname=0` to the user's exported `GOFLAGS` even if it's already present? If the script is run multiple times will it keep appending `-ldflags=-checklinkname=0`?

e2e/e2e_test.go
```diff
@@ -0,0 +1,151 @@
...
+	monomerClient := e2e.NewMonomerClient(monomerRPCClient)
+
+	bftClient, err := bftclient.New("tcp://127.0.0.1:26657", "/websocket")
+	require.NoError(t, err)
+	require.NoError(t, bftClient.Start())
+
+	{
+		engineurl, err := url.ParseString("ws://127.0.0.1:9001")
+		require.NoError(t, err)
+		engineurl.IsReachable(context.Background())
```
natebeauregard, line 110: Do we need to assert that the engineurl is reachable in both of these places?

`require.True(t, engineurl.IsReachable(context.Background()))`

e2e/stack_test.go
```diff
@@ -264,23 +121,35 @@ func ethRollupFlow(t *testing.T, stack *e2e.StackConfig) {
 	balanceBeforeDeposit, err := l1Client.BalanceAt(stack.Ctx, userETHAddress, nil)
 	require.NoError(t, err)
 
+	// wait for two blocks
+	require.NoError(t, stack.WaitL1(2))
```
natebeauregard, line 125: curious - why do you need to wait for two blocks here? Is this to prevent the underflow error you were seeing earlier?

e2e/stack_test.go
```diff
@@ -264,23 +121,35 @@ func ethRollupFlow(t *testing.T, stack *e2e.StackConfig) {
...
+	// https://github.com/ethereum-optimism/optimism/blob/24a8d3e06e61c7a8938dfb7a591345a437036381/op-e2e/tx_helper.go#L38
+	depositTx, err := PadGasEstimate(
+		createL1TransactOpts(t, stack, userPrivKey, l1signer, depositAmount),
+		1.1,
+		func(opts *bind.TransactOpts) (*types.Transaction, error) {
+			return stack.OptimismPortal.DepositTransaction(
+				opts,
+				common.Address(userCosmosETHAddress),
+				depositAmount,
+				100_000,  // 0l2GasLimit, // l2GasLimit,
```
natebeauregard, line 138: nit
```suggestion
				100_000,  // l2GasLimit,
```

e2e/stack_test.go
```diff
@@ -317,52 +186,70 @@ func ethRollupFlow(t *testing.T, stack *e2e.StackConfig) {
...
 	// inspect L2 events to ensure that the user's ETH was burned on L2
-	requireEthIsBurned(t, stack, userCosmosAddr, depositValueHex)
+	requireEthIsBurned(t, stack, userCosmosAddr, hexutil.EncodeBig(withdrawalAmount))
 
 	// wait for the L2 output containing the withdrawal tx to be proposed on L1
-	l2OutputBlockNumber := waitForL2OutputProposal(t, stack.L2OutputOracleCaller)
+	var l2OutputBlockNumber *big.Int
+	for range 5 {
+		l2OutputBlockNumber = waitForL2OutputProposal(t, stack.L2OutputOracleCaller)
+	}
```
natebeauregard, line 244: Why do we need to wait for five output proposals?

e2e/stack_test.go
```diff
@@ -317,52 +186,70 @@ func ethRollupFlow(t *testing.T, stack *e2e.StackConfig) {
...
-	withdrawalTx := e2e.NewWithdrawalTx(0, common.Address(userCosmosETHAddress), userETHAddress, depositAmount, new(big.Int).SetUint64(params.TxGas))
+	withdrawalTx := e2e.NewWithdrawalTx(0, common.Address(userCosmosETHAddress), userETHAddress, withdrawalAmount, new(big.Int).SetUint64(params.TxGas))
 
-	// initiate the withdrawal of the deposited amount on L2
-	senderAddr, err := monomer.CosmosETHAddress(*withdrawalTx.Sender).Encode("cosmos")
+	queryAccountBytes, err := protov1.Marshal(&authv1beta1.QueryAccountRequest{
+		Address: userCosmosAddr,
+	})
+	require.NoError(t, err)
+	result, err := stack.L2Client.ABCIQuery(stack.Ctx, authv1beta1.Query_Account_FullMethodName, queryAccountBytes)
```
natebeauregard, line 208: nit: `result` may seems like a pretty generic var name, could you rename it to something like `queryResult` for readability?

e2e/stack_test.go
```diff
@@ -396,17 +283,11 @@ func ethRollupFlow(t *testing.T, stack *e2e.StackConfig) {
...
 
 	// wait for the withdrawal finalization period before sending the withdrawal finalizing tx
-	finalizationPeriod, err := stack.L2OutputOracleCaller.FinalizationPeriodSeconds(&bind.CallOpts{})
-	require.NoError(t, err)
-	time.Sleep(time.Duration(finalizationPeriod.Uint64()) * time.Second)
-
-	// get the user's balance before the withdrawal has been finalized
-	balanceBeforeFinalization, err := stack.L1Client.BalanceAt(stack.Ctx, userETHAddress, nil)
-	require.NoError(t, err)
+	require.NoError(t, stack.WaitL1(2)) // TODO why do we need to wait this long?
```
natebeauregard, line 286: I'm curious as well, I left a comment above asking a similar question about waiting two blocks on L1. I see that we [also wait two blocks on L2 above](https://github.com/polymerdao/monomer/pull/267/files#diff-bad29d90d386ba872f7b623291e3e7658548ce5f925bf5fe5b34b05ed8ab3da6R235) as well.

e2e/stack_test.go
```diff
@@ -600,38 +535,20 @@ func createL1TransactOpts(
...
-	address := crypto.PubkeyToAddress(user.PublicKey)
 	return &bind.TransactOpts{
-		From: address,
+		From: ethcrypto.PubkeyToAddress(user.PublicKey),
 		Signer: func(addr common.Address, tx *types.Transaction) (*types.Transaction, error) {
 			signed, err := types.SignTx(tx, l1signer, user)
 			require.NoError(t, err)
 			return signed, nil
 		},
-		Nonce:    getCurrentUserNonce(t, stack, address),
```
natebeauregard, line 614: Nice that we can remove explicitly setting the nonce! Is this inferred somewhere by default for us?

integrations/integrations.go
```diff
@@ -238,10 +238,19 @@ func startOPDevnet(
 	} else {
 		deployConfig.L1GenesisBlockTimestamp = hexutil.Uint64(time)
 	}
-	l1Allocs, err := readFromFileOrGetDefault(v.GetString(flagL1AllocsPath), opdevnet.DefaultL1Allocs)
+	// TODO thinking about removing this clunky readFromFileOrGetDefault abstraction.
```
natebeauregard, line 241: I think its fine to leave as is for now. If you feel like we should eventually update this lmk and I can make a new issue for it.

natebeauregard (approved):
Changes look great! Left a few minor comments and questions but nothing blocking

e2e/stack_test.go
joshklop: Yes I believe so. I don't remember the exact flow, but it works so I assume the answer is yes.

integrations/integrations.go
joshklop: Agreed. We will need to revamp the integrations package when using server/v2 I believe, so probably best to leave it be.

e2e/stack_test.go
joshklop: Waiting for one wasn't enough. I don't understand the root cause. Would be good to understand this.

e2e/e2e_test.go
joshklop: Ah yeah forgot to refactor this. Thanks

cmd/monogen/testapp/setup-helper.sh
joshklop: Actually, we probably don't need to worry about these things.

Key idea: environment variables are copied into child processes.

> Will this append -ldflags=-checklinkname=0 to the user's exported GOFLAGS even if it's already present?

No, the script it will only modify the env var in its own process (not its parent's process).

> If the script is run multiple times will it keep appending -ldflags=-checklinkname=0?

No, since the script will take the `GOFLAGS` var from the parent process and only modify it within its own process.

e2e/stack_test.go
```diff
@@ -433,50 +365,104 @@ func ethRollupFlow(t *testing.T, stack *e2e.StackConfig) {
 	require.True(t, finalizeWithdrawalLogs.Event.Success, "withdrawal finalization failed")
 	require.NoError(t, finalizeWithdrawalLogs.Close())
 
-	// get the user's balance after the withdrawal has been finalized
-	balanceAfterFinalization, err := stack.L1Client.BalanceAt(stack.Ctx, userETHAddress, nil)
```
natebeauregard, line 437: Should these balance checks to ensure that the L1 recipient address got the withdrawn funds be added back in?

joshklop: Those only check that the gas for the L1 transaction was deducted, correct? My understanding is that they effectively test the L1 full node, not Monomer.

PR #270, opened 2024-10-25
Add L1 Data Fee Deduction

### LINKED ISSUE #268: Add a new decorator for L1 data fees
**Overview**
Currently, Monomer applications will contain all execution gas fees in the `FeeCollector` module account.

However, we need to add gas consumption for L1 data fees as well to fulfill the OP Stack Spec and compensate the sequencer for posting data on L1. To do this, we can add a new decorator to the default `AnteHandler` to calculate the L1 data fee. For posting batches via blobs, we can use the following [formula](https://specs.optimism.io/protocol/exec-engine.html#ecotone-l1-cost-fee-changes-eip-4844-da):

`(compressedTxSize) * (16*l1BaseFee*lBaseFeeScalar + l1BlobBaseFee*l1BlobBaseFeeScalar) / 1e6`

**Tasks**
- Update monogen to use a custom `AnteHandler` for regular transactions.
    - We’re currently using a custom `AnteHandler` to bypass signatures and fees for `Deposit` txs, however we’ll want to use a custom one for regular txs as well to add an additional decorator to the default `x/auth` `AnteHandler`
- Add a new decorator (`ConsumeGasForL1DataFeeDecorator`) to the `AnteHandler` to include cosuming gas for L1 data fees.
    - Can be based on `ConsumeGasForTxSizeDecorator`
    - Use [Ecotone](https://specs.optimism.io/protocol/exec-engine.html#ecotone-l1-cost-fee-changes-eip-4844-da) calculation for blob support.

natebeauregard:
Closes #268 

**Overview**
Currently, Monomer applications will contain all execution gas fees in the `FeeCollector` module account.

However, we need to add gas consumption for L1 data fees as well to fulfill the OP Stack Spec and compensate the sequencer for posting data on L1. To do this, we can add a new decorator to the default `AnteHandler` to calculate the L1 data fee. For posting batches via blobs, we can use the following [formula](https://specs.optimism.io/protocol/exec-engine.html#ecotone-l1-cost-fee-changes-eip-4844-da):

`(zeroes*4 + ones*16) * (16*l1BaseFee*l1BaseFeeScalar + l1BlobBaseFee*l1BlobBaseFeeScalar) / 16e6`

**Implementation**
- Adds a new decorator (`ConsumeGasForL1DataDecorator`) to the Monomer `AnteHandler` to include cosuming gas for L1 data fees.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method to retrieve Layer 1 block information, enhancing transaction processing capabilities.
	- Added functionality to handle Layer 1 data transactions, including gas cost estimation for Ethereum postings.

- **Improvements**
	- Enhanced the `AnteHandler` to integrate rollup operations, improving transaction handling flow.
	- Updated error handling for various operations related to transaction processing and block information retrieval, ensuring clearer feedback on issues.
	- Improved clarity in method comments for better understanding of functionalities.


x/rollup/tx/helpers/l1_data_decorator.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+}
+
+func (d ConsumeGasForL1DataDecorator) AnteHandle(
+	ctx sdk.Context, //nolint:gocritic // hugeparam
+	tx sdk.Tx,
+	simulate bool,
+	next sdk.AnteHandler,
+) (sdk.Context, error) {
+	l1BlockInfo, err := d.rollupKeeper.GetL1BlockInfo(ctx)
```
natebeauregard, line 32: I looked into adding unit/integration test coverage for a while and eventually decided to split it out for the time being since it was blowing up in scope.

I created https://github.com/polymerdao/monomer/issues/281 to refactor how we pass around the rollup keeper to use an interface that can be mocked instead of a struct and to add unit tests for `ante.go` and `l1_data_decorator.go`.

Also, this logic will eventually be tested in our e2e tests once https://github.com/polymerdao/monomer/pull/267 merges.

x/rollup/tx/helpers/l1_data_ante.go
joshklop, line 1: can we make the `ConsumeGasForL1DataDecorator` an ante handler and then remove this file? Seems simpler and removes some indirection. Let me know what you think.

x/rollup/tx/helpers/ante.go
```diff
@@ -3,22 +3,57 @@ package helpers
...
+type HandlerOptions struct {
+	AccountKeeper          authante.AccountKeeper
+	BankKeeper             authtypes.BankKeeper
+	RollupKeeper           *rollupkeeper.Keeper
+	ExtensionOptionChecker authante.ExtensionOptionChecker
+	FeegrantKeeper         authante.FeegrantKeeper
+	SignModeHandler        *txsigning.HandlerMap
+	SigGasConsumer         func(meter storetypes.GasMeter, sig signing.SignatureV2, params authtypes.Params) error
+	TxFeeChecker           authante.TxFeeChecker
+}
```
joshklop, line 33: Rather than modifying the options. Can we just pass in the rollup keeper as a separate argument?

x/rollup/tx/helpers/l1_data_decorator.go
```diff
@@ -0,0 +1,91 @@
...
+
+	l1GasUsed, err := getL1GasUsed(tx)
+	if err != nil {
+		return ctx, fmt.Errorf("get l1 gas used: %w", err)
+	}
+
+	const baseFeeMultiplier = 16
+	const divisor = 16e6
+
+	// scaledBaseFee calculation: scaledBaseFee = l1BaseFee * lBaseFeeScalar * 16
```
joshklop, line 50: very nice comments, thanks!

x/rollup/tx/helpers/l1_data_decorator.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// scaledBlobBaseFee calculation: scaledBlobBaseFee = l1BlobBaseFee * l1BlobBaseFeeScalar
+	scaledBlobBaseFee := new(big.Int).Mul(blobBaseFee, blobBaseFeeScalar)
+
+	// l1DataCost calculation: l1DataCost = [l1GasUsed * (scaledBaseFee + scaledBlobBaseFee)] / 16e6
+	l1DataCost := new(big.Int).Div(new(big.Int).Mul(l1GasUsed, new(big.Int).Add(scaledBaseFee, scaledBlobBaseFee)), big.NewInt(divisor))
+
+	if !l1DataCost.IsUint64() {
+		return ctx, fmt.Errorf("l1 data cost overflow: %s", l1DataCost)
+	}
```
joshklop, line 61: is this a possibility or is it technically impossible?

natebeauregard: In practice I don't think that this block should ever be hit, however I wanted to add the extra safeguard since we're dealing with a lot of multiplication here. This would prevent against the case where one of the L1 attributes either unexpectedly increased or was misconfigured in the system deposit tx.

That being said, I'm not sure how useful this check is. Do you think I should just remove this block for the overflow case since it shouldn't ever be hit in production?

cmd/monogen/testapp/app/app.go
```diff
@@ -167,7 +167,7 @@ func New(
 		AccountKeeper:   app.AccountKeeper,
 		BankKeeper:      app.BankKeeper,
 		SignModeHandler: app.txConfig.SignModeHandler(),
-	})
+	}, app.RollupKeeper)
```
natebeauregard, line 170: Posting for visibility - we'll need to cut a release after this change so the monogen tool can pick up the ante handler updates.

cc @joshklop

joshklop: Monogen is pinned to the `v0.1.1` release, so it isn't a necessity. Am I missing something?

x/rollup/tx/l1_data.go
```diff
@@ -0,0 +1,81 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/ethereum/go-ethereum/rlp"
+	rollupkeeper "github.com/polymerdao/monomer/x/rollup/keeper"
+)
+
+// L1DataAnteHandler will consume gas to compensate the sequencer for posting
+// the transaction to Ethereum. The gas cost is calculated based on the Ecotone
+// upgrade and the sequencer is expected to post the transaction using blobs.
+func L1DataAnteHandler(ctx sdk.Context, tx sdk.Tx, rollupKeeper *rollupkeeper.Keeper) (sdk.Context, error) { //nolint:gocritic // hugeparam
```
joshklop, line 17: Should the behavior change if the tx is being simulated? We currently don't accept a `simulate` arg like other ante handlers, not sure if that's an issue.

x/rollup/tx/l1_data.go
```diff
@@ -0,0 +1,81 @@
...
+	blobBaseFee := new(big.Int).SetBytes(l1BlockInfo.BlobBaseFee)
+	blobBaseFeeScalar := big.NewInt(int64(l1BlockInfo.BlobBaseFeeScalar))
+
+	l1GasUsed, err := getL1GasUsed(tx)
+	if err != nil {
+		return ctx, fmt.Errorf("get l1 gas used: %w", err)
+	}
+
+	const baseFeeMultiplier = 16
+	const divisor = 16000000 // 16e6
```
joshklop, line 38: nit: I think this should work.
```suggestion
	const divisor = 16e6
```

joshklop (approved):

x/rollup/tx/l1_data.go
natebeauregard: I set `16e6` directly in the constant before but I just added a new [commit](https://github.com/polymerdao/monomer/pull/270/commits/db7bef55cc6bca3caa68ce8fc8eba8c906f2c4df) a couple hours ago to modify this to be an int instead since 16e6 defaults to a float64 type. Idt it makes a huge difference though so I'll go ahead and change it back

x/rollup/tx/l1_data.go
natebeauregard: There shouldn't be any behavior difference for us during `simulate` mode since we're basing the gas consumption on the current L1 attributes and the size of the tx. The current L1 attributes could change between the simulated tx and the real tx and result in an out-of-gas error. However, the user is expected to account for that when setting their gas limit for the real tx (normally 1.1x or 1.2x of the gas used for the simulated tx).

 Also, `simulate` is passed into `DepositAnteHandler` but it's never used there.

cmd/monogen/testapp/app/app.go
natebeauregard: Yeah it's not a necessity, however `monogen` will be broken on main when testing the current monomer changes. I think your post in the discord thread made sense and this should be covered in the weekly releases.

PR #273, opened 2024-10-27
Derive Cosmos addresses from public keys

joshklop:


joshklop:
I was planning to do this later, but it's required for fixing the e2e tests.

joshklop:
Now that I think about it, I believe the [address conversions](https://github.com/polymerdao/monomer/blob/63bbf2c25b07533ca6a224f0cc84c09adac4adf0/cmd/monogen/testapp/wallet/metamask-integration.js#L115-L130) in the `wallet` web app will need to change. We'll want to simply obtain the data part of the bech32 encoding, which should be simple, but I'm not sure how to do that using the `bech32` js library yet.

natebeauregard (commented):
Monomer changes LGTM. I'll wait to approve until the frontend wallet server changes are in as well. I wonder if we can use some of the helpers from the [Evmos JS converter](https://github.com/evmos/evmosjs/blob/main/packages/address-converter/src/converter.ts).

joshklop:
Hmm, it seems to work fine for me. Perhaps I'm doing something wrong though. Mind checking on your end @natebeauregard ?

natebeauregard:
> Hmm, it seems to work fine for me. Perhaps I'm doing something wrong though. Mind checking on your end @natebeauregard ?

When you say `it seems to work fine for me` are you referring to the frontend `bech32` library in the wallet integration server? If so that's great, have you tested with prefixes other than "cosmos"? I can test as well on my end once I get some more context

joshklop:
> When you say it seems to work fine for me are you referring to the frontend bech32 library in the wallet integration server? 

I'm able to deposit funds to L2 and transfer them around. I would expect that the different address translation would cause them to be sent to a different address (i.e., not my keplr address). However, thinking about this again, it actually makes sense: the way you did address translation in frontend happened to be compatible with both address conversion methods.

> If so that's great, have you tested with prefixes other than "cosmos"?

I have not. Might be interesting to try it though. I'll see if I can get to it tomorrow, pretty unlikely however.

natebeauregard:
> > When you say it seems to work fine for me are you referring to the frontend bech32 library in the wallet integration server? 
> 
> I'm able to deposit funds to L2 and transfer them around. I would expect that the different address translation would cause them to be sent to a different address (i.e., not my keplr address). However, thinking about this again, it actually makes sense: the way you did address translation in frontend happened to be compatible with both address conversion methods.
> 
> > If so that's great, have you tested with prefixes other than "cosmos"?
> 
> I have not. Might be interesting to try it though. I'll see if I can get to it tomorrow, pretty unlikely however.
> 

Sounds good, thanks for the clarification. I can test the prefix case tm morning and can approve/merge if it works!

Issue #274, opened 2024-10-27
Add HTML, CSS, JavaScript linter


joshklop:
See the `wallet` directory in the testapp.

Nandini1071:
Hi @joshklop is this still open, if yes could you tell the exact requirement would like to contribute.

joshklop:
@Nandini1071 Something like a prettier config and integration into the `make lint` Makefile target would probably be sufficient.

PR #275, opened 2024-10-27
Automatically switch wallet network

joshklop:
## Summary by CodeRabbit

- **New Features**
	- Enhanced MetaMask integration for a more streamlined connection process.
	- Improved error handling for a better user experience when connecting to Ethereum chains.

- **Bug Fixes**
	- Removed the button for adding the L1 chain to MetaMask, which may affect previous user interactions.

- **Refactor**
	- Introduced a constant for `chainId` to improve code maintainability and consistency. 
	- Updated connection process to be initiated automatically on page load.
	- Added a new instruction to reload the page after configuring the Monomer L1 Devnet wallet.


cmd/monogen/testapp/wallet/metamask-integration.js
natebeauregard, line 1: Do you need to re-zip the testapp so it includes the latest wallet server changes?

cmd/monogen/testapp/wallet/metamask-integration.js
```diff
@@ -9,9 +9,9 @@ const depositButton = document.getElementById('deposit-button');
 const statusParagraph = document.getElementById('status');
 const amountInput = document.getElementById('amount');
 const recipientInput = document.getElementById('recipient');
-const enableMetamaskButton = document.getElementById('enable-metamask');
+const chainId = "0x384";
```
natebeauregard, line 12: nit: could you re-add the `// 900` comment so the `chainID` is more readable? It's kind of unfortunate that we need to pass through a hex value for chain ID when it's always represented as a base 10 number to the user

cmd/monogen/testapp/wallet/metamask-integration.js
natebeauregard, line 1: I know you mentioned something about using the Coinbase Wallet (or other alternative L1 wallets) as well in the discord thread. Were you able to verify that the wallet integration server works for other L1 wallets? IIRC the `window.ethereum` may be unique to the MetaMask wallet so I'm not sure if it would work, however that would be awesome if it did

cmd/monogen/testapp/wallet/index.html
```diff
@@ -129,7 +129,6 @@ <h2>Metamask Integration</h2>
                         </ul>
                     </li>
                 </ul>
-                <button id="enable-metamask">Add L1 Chain to Metamask</button>
```
natebeauregard, line 132: We'll probably want to update the `Metamask Integration` docs above with a step to reload the page once the wallet was removed if it was previously configured.

cmd/monogen/testapp/wallet/metamask-integration.js
natebeauregard, line 1: It would be nice to eventually include this change for the L2 wallet as well so we don't need to manually re-add it with the `Add L2 Chain to Keplr` button. No need to include that in this PR but we should do it eventually for parity.

joshklop: It does work. From [ERC 1193](https://eips.ethereum.org/EIPS/eip-1193)
> Historically, Providers have been made available as window.ethereum in web browsers, but this convention is not part of the specification.

cmd/monogen/testapp/wallet/index.html
joshklop: > if it was previously configured

What do you mean by this?

joshklop: I think I see now.

cmd/monogen/testapp/wallet/metamask-integration.js
joshklop: I went ahead and did it this time. However, it's really only necessary when we want to do a new `monogen` release.

joshklop: https://github.com/polymerdao/monomer/pull/277

natebeauregard (approved):

cmd/monogen/testapp/wallet/metamask-integration.js
natebeauregard, line 1: Should this file be renamed to something like `l1-integration.js` since it's not metamask specific?

joshklop: We'll put the call to the keplr integration in there eventually #277, so `l1-integration.js` will be misleading too. Probably do a full refactor once we have linting #274.

PR #277, opened 2024-10-28
Auto switch l2 network

joshklop:


cmd/monogen/testapp/wallet/metamask-integration.js
```diff
@@ -30,18 +30,37 @@ async function connectMetamask() {
             signer
         );
 
-        console.log('Connected to MetaMask and contract instantiated.');
-    } catch (error) {
-        console.error('Error connecting to MetaMask:', error);
-    }
+    await addL1ChainToMetaMask();
+    await switchToMonomerDevnet();
+    await runKeplrIntegration();
```
natebeauregard, line 35: If we're including the keplr integration logic in this file as well should we just move all of the related funcs over from the `keplr-integration.js` file and remove that file?

Alternatively, can we remove the following [block](https://github.com/polymerdao/monomer/blob/main/cmd/monogen/testapp/wallet/keplr-integration.js#L84-L85) at the end of the `keplr-integration.js` file since it no longer applies?
```
const enableKeplrButton = document.getElementById("enable-keplr");
enableKeplrButton.addEventListener('click', runKeplrIntegration);
```

natebeauregard:
> @joshklop While testing custom prefixes for #273 I realized that we'll probably want the step to add the L2 chain config to still be a button rather than having it immediately add on page load. If a user is using a custom prefix then the `bech32Config` that's passed to keplr will need to be updated with the expected bech32 prefixes.
> 
> We'll probably need the user to add their prefix through a text input field and then we can update the bech32 config for them whenever they click the button to add their chain to Keplr.

For visibility - Talked about this in the daily meeting and we won't include the custom prefix change in this PR. That work will be included in #282

cmd/monogen/testapp/wallet/metamask-integration.js
joshklop: Do you mind if we merge this PR and do a proper refactor in a follow-up? We can tackle code organization and linting in a single PR I think.

natebeauregard: Yeah sure that sgtm

natebeauregard (approved):

PR #284, opened 2024-10-30
Manage go tools via tools.go

Jesserc:


Jesserc:
@joshklop

To manage Go tools using tools.go, we can follow these steps:

- Clone the repository.
- Run `go mod tidy`
- Install the necessary tools listed in tools.go (this runs `go install` for each tool):
	```bash
	cat tools.go | grep _ | awk -F'"' '{print $2}' | xargs -tI % scripts/go-wrapper.sh install %
	```
What are your thoughts on this approach?

Issue #269, opened 2024-10-24
Create a new `x/rollup` message type to withdraw collected fees to L1


natebeauregard:
**Overview**

To withdraw fees from L2 to L1, we can mimic the FeeVault.sol implementation in the Optimism repository in x/rollup. We can add a new message type in x/rollup to initiate a withdrawal from the FeeCollector account to the L1 recipient address once the FeeCollector account has more than a set amount of ETH.

**Tasks**
- Create a new `x/rollup` message type to initiate a withdrawal of funds from the `FeeCollector` account to the L1 recipient address.
    - When a fee withdrawal message is sent and the `FeeCollector` module account contains more ETH than the minimum withdrawal amount, a withdrawal initiating transaction can be sent.
        - This transaction can be sent permissionlessly since the L1 recipient is configured in `x/rollup` and it isn’t configurable by the message sender.
    - This will most likely share a good bit of code with the `MsgInitiateWithdrawal` flow, we should ensure we’re refactoring out and reusing shared code.

PR #287, opened 2024-10-31
Create a new x/rollup message type to withdraw collected fees to L1

natebeauregard:
Closes https://github.com/polymerdao/monomer/issues/269

Adds a message server implementation for the `InitiateFeeWithdrawal` message to mimic the behavior of the OPStack's FeeVault contracts.

Users can permissionlessly triggers fee withdrawals on L2 if the `FeeCollector` module account balance is above the predefined minimum withdrawal amount. The fees are then able to be withdrawn to the predefined target address on L1.

Also, updates the builder to use the withdrawal tx event attributes forconstructing the withdrawal hash instead of the withdrawal message.

The minimum withdrawal amount and L1 recipient address are currently hardcoded in `x/rollup`, however they'll be updated to be configurable once a `x/rollup` genesis state is added.

## Release Notes

- **New Features**
  - Introduced a new method for initiating fee withdrawals from Layer 2 to Layer 1.
  - Enhanced withdrawal message handling to include fee withdrawals.

- **Improvements**
  - Streamlined the `InitiateWithdrawal` method by consolidating parameters into a single struct.
  - Updated transaction handling for better clarity and precision in encoding amounts.
  - Added a new constant for Layer 2 fee withdrawal transactions.
  - Enhanced the test suite for more comprehensive testing of fee-related operations.

- **Bug Fixes**
  - Added specific error handling for fee withdrawal failures.

- **Tests**
  - Expanded test coverage for the fee withdrawal process and improved mock setups for account management.


x/rollup/keeper/msg_server.go
```diff
@@ -83,3 +86,70 @@ func (k *Keeper) InitiateWithdrawal(
...
+	_ *types.MsgInitiateFeeWithdrawal,
+) (*types.MsgInitiateFeeWithdrawalResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// TODO: make minWithdrawalAmount and l1recipientAddr configurable once a genesis state is added
+	const (
+		minWithdrawalAmount   = 100
+		l1recipientAddr       = "0x63d93aC6FA6B4021527e967ac3Eb29F2B3E52B96"
+		feeWithdrawalGasLimit = 400_000
+	)
```
joshklop, line 101: Would it be easier to just make these params in the `MsgInitiateFeeWithdrawal`? Not necessary in this PR.

x/rollup/keeper/msg_server.go
```diff
@@ -83,3 +86,70 @@ func (k *Keeper) InitiateWithdrawal(
...
+
+	feeCollectorAddr := k.accountkeeper.GetModuleAddress(authtypes.FeeCollectorName)
+	if feeCollectorAddr == nil {
+		ctx.Logger().Error("Failed to get fee collector address")
+		return nil, types.WrapError(types.ErrInitiateFeeWithdrawal, "failed to get fee collector address")
```
joshklop, line 106: Is the standard pattern to log and then return an error? I feel like the error message should be logged further up.

x/rollup/keeper/msg_server.go
```diff
@@ -83,3 +86,70 @@ func (k *Keeper) InitiateWithdrawal(
...
+	}
+
+	feeCollectorBalance := k.bankkeeper.GetBalance(ctx, feeCollectorAddr, types.WEI)
+	if feeCollectorBalance.Amount.LT(math.NewInt(minWithdrawalAmount)) {
+		ctx.Logger().Error("Fee collector balance is below the minimum withdrawal amount", "balance", feeCollectorBalance.String())
```
joshklop, line 111: Do we know what the purpose of having a minimum is?

x/rollup/keeper/msg_server.go
```diff
@@ -83,3 +86,70 @@ func (k *Keeper) InitiateWithdrawal(
...
+			types.ErrInitiateFeeWithdrawal,
+			"fee collector balance is below the minimum withdrawal amount: %v", feeCollectorBalance.String(),
+		)
+	}
+
+	ctx.Logger().Debug("Withdrawing L2 fees", "amount", feeCollectorBalance.String(), "recipient", l1recipientAddr)
+
+	// Burn the withdrawn fees from the fee collector account on L2
+	fees := sdk.NewCoins(feeCollectorBalance)
+	if err := k.bankkeeper.SendCoinsFromModuleToModule(ctx, authtypes.FeeCollectorName, types.ModuleName, fees); err != nil {
```
joshklop, line 122: Is it necesssary to send them to the rollup module first?

x/rollup/types/expected_keepers.go
```diff
@@ -10,12 +10,14 @@ import (
...
 	MintCoins(ctx context.Context, name string, amt sdk.Coins) error
 	BurnCoins(ctx context.Context, moduleName string, amt sdk.Coins) error
+
+	GetBalance(ctx context.Context, addr sdk.AccAddress, denom string) sdk.Coin
 }
 
 type AccountKeeper interface {
-	NewAccountWithAddress(context.Context, sdk.AccAddress) sdk.AccountI
-	SetAccount(context.Context, sdk.AccountI)
+	GetModuleAddress(moduleName string) sdk.AccAddress
```
joshklop, line 22: Woops, was this an unused type before? I might have forgotten to remove it in a previous PR.

builder/builder.go
```diff
@@ -290,71 +291,107 @@ func (b *Builder) parseWithdrawalMessages(
...
 	// Initiate the withdrawal in the Monomer ethereum state.
-	if err = executer.InitiateWithdrawal(
-		common.BytesToAddress(senderCosmosAddress.Bytes()),
-		withdrawalMsg.Value.BigInt(),
-		common.HexToAddress(withdrawalMsg.GetTarget()),
-		new(big.Int).SetBytes(withdrawalMsg.GasLimit),
-		withdrawalMsg.GetData(),
-	); err != nil {
-		return nil, fmt.Errorf("initiate withdrawal: %v", err)
+	if err = executer.InitiateWithdrawal(*params); err != nil {
```
joshklop, line 347: Let's make `InitiateWithdrawal` accept a pointer to avoid copying (maybe the compiler is smart and I'm missing something, but we use pointers elsewhere anyway).

builder/builder.go
```diff
@@ -290,71 +291,107 @@ func (b *Builder) parseWithdrawalMessages(
...
+	var params crossdomain.Withdrawal
+	for _, attr := range withdrawalEvent.Attributes {
+		switch attr.Key {
+		case rolluptypes.AttributeKeySender:
+			senderCosmosAddress, err := sdk.AccAddressFromBech32(attr.Value)
+			if err != nil {
+				return nil, fmt.Errorf("convert sender to cosmos address: %v", err)
+			}
+			// TODO: use updated cosmos -> eth address conversion
+			sender := common.BytesToAddress(senderCosmosAddress.Bytes())
```
joshklop, line 371: Is this a TODO?

bindings/L2ToL1MessagePasserExecuter.go
```diff
@@ -35,6 +35,9 @@ func NewL2ToL1MessagePasserExecuter(evm *vm.EVM) (*L2ToL1MessagePasserExecuter,
 }
 
 func (e *L2ToL1MessagePasserExecuter) InitiateWithdrawal(params crossdomain.Withdrawal) error {
+	if params.Target == nil {
+		return fmt.Errorf("l1 target address is nil")
+	}
```
joshklop, line 40: Does `Target` need to be a pointer?

x/rollup/keeper/msg_server.go
natebeauregard: These params will eventually be module specific once a genesis state is added instead of stored as constants.

However, they can't be added as a field to `MsgInitiateFeeWithdrawal` since any L2 user can permissionlessly trigger a fee withdrawal. If these were fields on `MsgInitiateFeeWithdrawal`, then they could maliciously forward the fees to a custom L1 recipient address instead of the predefined L1 address that the sequencer expects.

x/rollup/keeper/msg_server.go
natebeauregard: Yeah I agree that this pattern is pretty confusing and not too helpful, especially since we haven't gotten any user feedback on debugging Monomer L2s yet.

I think the initial `ctx.Logger()...` pattern before the SDK error was inherited code and I decided to just keep it going. I'm leaning towards removing all of the `ctx.Logger()...` calls in `x/rollup` until we do some usability testing, is that alright with you?

x/rollup/keeper/msg_server.go
natebeauregard: The [spec](https://specs.optimism.io/protocol/predeploys.html#sequencerfeevault) is kind of vague on the purpose of the minimum withdrawal amount:
> When enough fees accumulate in this account, they can be withdrawn to an immutable L1 address.

Since any L2 user can initiate a fee withdrawal, I'm guessing it's a safeguard to prevent a potential attack where an L2 user could spam fee withdrawals and then the sequencer would be paying more in gas on L1 than they would receive in fees.

x/rollup/keeper/msg_server.go
natebeauregard: Not necessarily - the alternative here would be to have the app enable burn permissions to the `FeeCollector` account, however I'm not entirely sure if that could open the door for any additional unwanted behavior.

Since we already require the user to enable burn permissions for the `x/rollup` module for user withdrawals, I decided to just forward the fees to the rollup module first before burning the fees.

Lmk what your thoughts on enabling burn permissions for the `FeeCollector` module account are and if you think that would be a better approach.

x/rollup/types/expected_keepers.go
natebeauregard: I didn't look too much into when they were added and if we were previously using them, however they weren't being used on main so I just decided to remove them.

builder/builder.go
natebeauregard: Yeah good call, I'll update this

builder/builder.go
natebeauregard: Ah man yeah I forgot to remove the TODO, thanks for pointing this out. I'll remove it

bindings/L2ToL1MessagePasserExecuter.go
natebeauregard: Probably not, I'll look into this and update it if a pointer isn't necessary

x/rollup/keeper/msg_server.go
joshklop: Let's ensure the errors get logged somewhere and remove if they do get logged

x/rollup/keeper/msg_server.go
joshklop: This sounds reasonable. Mind putting that in a comment?

joshklop (approved):

PR #289, opened 2024-11-05
Add x/rollup genesis state

natebeauregard:
Adds a genesis state to x/rollup to allow module specific params to be configured.

The following module specific params were added to the `x/rollup` genesis state:
-  **l1_fee_recipient**: L1 address to forward the L2 fees to.
-  **l1_cross_domain_messenger**: L1 address of the cross-domain messenger contract.
-  **min_fee_withdrawal_amount**: Minimum amount of L2 fees that the FeeCollector account must have before they can be withdrawn.
-  **fee_withdrawal_gas_limit**: L1 gas limit for withdrawing fees to the L1 recipient address.


x/rollup/types/genesis.go
```diff
@@ -0,0 +1,22 @@
...
+// DefaultGenesisState - Return a default genesis state
+func DefaultGenesisState() *GenesisState {
+	return NewGenesisState(DefaultParams())
+}
+
+// ValidateGenesis performs basic validation of rollup genesis data returning an
+// error for any failed validation criteria.
+func ValidateGenesis(state GenesisState) error {
+	if err := state.Params.Validate(); err != nil {
+		return err
```
joshklop, line 19:
```suggestion
		return fmt.Errorf("validate params: %v", err)
```

x/rollup/types/params.go
```diff
@@ -0,0 +1,50 @@
...
+
+// NewParams creates a new Params object
+func NewParams(l1FeeRecipient, l1CrossDomainMessenger string, minFeeWithdrawalAmount, feeWithdrawalGasLimit uint64) Params {
+	return Params{
+		L1FeeRecipient:         l1FeeRecipient,
+		L1CrossDomainMessenger: l1CrossDomainMessenger,
+		MinFeeWithdrawalAmount: minFeeWithdrawalAmount,
+		FeeWithdrawalGasLimit:  feeWithdrawalGasLimit,
+	}
+}
```
joshklop, line 26: Can we get rid of this `NewParams` function since there is no validation and all fields are public?

x/rollup/types/params.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// Default parameter values
+const (
+	DefaultL1FeeRecipient string = "0x000000000000000000000000000000000000dEaD"
+	// DefaultL1CrossDomainMessenger uses the devnet address of the L1 cross domain messenger contract as the default value.
+	DefaultL1CrossDomainMessenger string = "0x9A9f2CCfdE556A7E9Ff0848998Aa4a0CFD8863AE"
+	DefaultMinFeeWithdrawalAmount uint64 = 400_000
+	DefaultFeeWithdrawalGasLimit  uint64 = 400_000
```
joshklop, line 15: I'd prefer to delete these global vars, and only have one `DefaultParams()` function that returns a `Params` object.

x/rollup/types/params.go
```diff
@@ -0,0 +1,50 @@
...
+	return NewParams(DefaultL1FeeRecipient, DefaultL1CrossDomainMessenger, DefaultMinFeeWithdrawalAmount, DefaultFeeWithdrawalGasLimit)
+}
+
+// Validate checks that the parameters have valid values.
+func (p *Params) Validate() error {
+	if err := validateEthAddress(p.L1FeeRecipient); err != nil {
+		return fmt.Errorf("invalid L1 fee recipient address: %w", err)
+	}
+	if err := validateEthAddress(p.L1CrossDomainMessenger); err != nil {
+		return fmt.Errorf("invalid L1 cross domain messenger address: %w", err)
```
joshklop, line 39:
```suggestion
		return fmt.Errorf("validate L1 cross domain messenger address: %v", err)
```

x/rollup/keeper/genesis.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/polymerdao/monomer/x/rollup/types"
+)
+
+// InitGenesis - Init store state from genesis data
+func (k *Keeper) InitGenesis(ctx sdk.Context, state types.GenesisState) { //nolint:gocritic // hugeParam
```
joshklop, line 11: can we return errors here? Same for `ExportGenesis`

joshklop (approved):
Looks great, just a few small suggestions.

x/rollup/keeper/genesis.go
natebeauregard: We could add logging here to add additional messenging for the user, however we need to panic instead of throwing an error to properly implement the Cosmos SDK module `HasGenesis`. Unfortunately this is how other Cosmos SDK modules handle their `InitGenesis` and `ExportGenesis` cases.

joshklop: Got it, no worries then.

PR #310, opened 2024-12-05
Only allow deposit messages from Monomer

joshklop:
Before, we were only blocking deposit transactions from the Monomer mempool. This meant that standard auth txs could contain deposit messages, minting arbitrary amounts of eth. We fix this by checking for deposit messages in all transactions in helpers/ante.go.

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for transactions containing deposit messages.
	- Introduced a new `DepositMsg` interface for better message type management.
	- Streamlined transaction creation process with new utility functions.

- **Bug Fixes**
	- Removed the handling of deposit transactions from the enqueueing process, simplifying transaction management.

- **Tests**
	- Added unit tests for the ante handler to ensure transactions with deposit messages are correctly rejected.
	- Simplified test suite by removing the test case for deposit transactions while maintaining validation for other transaction behaviors.

### DIFF
--- a/e2e/stack_test.go
+++ b/e2e/stack_test.go
@@ -8,17 +8,10 @@ import (
 	"time"
 
 	authv1beta1 "cosmossdk.io/api/cosmos/auth/v1beta1"
-	txv1beta1 "cosmossdk.io/api/cosmos/tx/v1beta1"
 	"cosmossdk.io/math"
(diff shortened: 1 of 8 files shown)


natebeauregard (approved):
LGTM!

PR #312, opened 2024-12-06
ERC-20 Withdrawals

natebeauregard:
Add a message type to initiate ERC-20 withdrawals on L2. The ERC-20 token will be burned on L2 and will encode messages to send to the L1CrossDomainMessenger and L1StandardBridge to release the deposited ERC-20 token on L1 after the withdrawal is successfully proven and finalized.

## Release Notes

- **New Features**
  - Introduced support for ERC-20 token withdrawals, allowing users to initiate withdrawals of ERC-20 tokens alongside ETH.
  - Added a new parameter for the L1 standard bridge in the configuration options.

- **Enhancements**
  - Improved error handling and validation for withdrawal messages, including checks for valid Ethereum addresses.
  - Enhanced logging for withdrawal processes to provide better insights during transactions.
  - Updated message types and parameters to improve clarity and functionality in withdrawal processes.

- **Bug Fixes**
  - Corrected error messages for better clarity in withdrawal operations.

- **Tests**
  - Expanded test coverage for both ETH and ERC-20 withdrawals, ensuring robust validation and error handling.


x/rollup/keeper/msg_server_test.go
```diff
@@ -129,6 +129,7 @@ func (s *KeeperTestSuite) TestApplyUserDeposit() {
 func (s *KeeperTestSuite) TestInitiateWithdrawal() {
 	sender := sdk.AccAddress("addr").String()
 	l1Target := "0x12345abcde"
+	erc20TokenAddress := "0x0123456789abcdef"
```
joshklop, line 132: Should we have a separate test for erc20 withdrawals? Interspersing the two seems looks a bit complicated.

x/rollup/types/msgs.go
```diff
@@ -74,9 +74,39 @@ func (m *MsgInitiateWithdrawal) ValidateBasic() error {
...
+	return "l2_eth_withdrawal"
 }
 
 func (*MsgInitiateWithdrawal) Route() string {
 	return ModuleName
 }
+
+var _ sdktypes.Msg = (*MsgInitiateERC20Withdrawal)(nil)
+
+func (m *MsgInitiateERC20Withdrawal) ValidateBasic() error {
```
joshklop, line 86: ~Do you know where these `ValidateBasic`s are called?~ [Never mind](https://github.com/cosmos/cosmos-sdk/blob/8bfcf554275c1efbb42666cc8510d2da139b67fa/baseapp/baseapp.go#L879)

joshklop (approved):

x/rollup/keeper/msg_server_test.go
natebeauregard: Yeah I'll split them out, I kept going back and forth on this but I agree that it would be a lot more readable if they were split even if a lot of the same code is shared

builder/builder.go
```diff
@@ -362,11 +363,17 @@ func parseWithdrawalEventAttributes(withdrawalEvent *abcitypes.Event) (*crossdom
 	for _, attr := range withdrawalEvent.Attributes {
 		switch attr.Key {
 		case rolluptypes.AttributeKeySender:
-			senderCosmosAddress, err := sdk.AccAddressFromBech32(attr.Value)
-			if err != nil {
-				return nil, fmt.Errorf("convert sender to cosmos address: %v", err)
+			// check whether the sender address should be encoded as an Ethereum address or a Cosmos address
+			var sender common.Address
+			if common.IsHexAddress(attr.Value) {
+				sender = common.HexToAddress(attr.Value)
```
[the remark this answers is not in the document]
natebeauregard: For regular ETH withdrawals, we just use the Cosmos (bech32) address that triggered the withdrawal on L2 as the sender field in the withdrawal tx for proving and finalizing on L1. This is due to us not having the public key associated with that cosmos ETH address, however it shouldn't pose any issues since sender field is only used to calculate the withdrawal hash and is then emitted in an event on L1.

To successfully prove/finalize ERC-20 withdrawals on L1, we need some extra assertions that the sender field for the withdrawal is the [L2CrossDomainMessenger](https://github.com/polymerdao/monomer/blob/nkb.erc-20-withdrawals/x/rollup/keeper/msg_server.go#L157-L158) address. Since this address is static and known at genesis, we don't need to worry about any public key conversions and can just pass through the expected L2CrossDomainMessenger address. The extra check added in the builder is to verify that both types of addresses (cosmos bech32, ETH hex) can be parsed as expected.

PR #320, opened 2024-12-16
Enable starting monomer apps in verifier mode

natebeauregard:
Introduces the `--monomer.sequencer-mode` flag to start the Monomer application in sequencer mode. Otherwise, the Monomer application will be started in verifier mode.

## Summary by CodeRabbit

- **Documentation**
	- Updated instructions for creating a Monomer application, clarifying command-line options for running in sequencer and verifier modes.
	- Revised documentation for interacting with Monomer Rollup Devnet to reflect changes in the startup command.
- **New Features**
	- Introduced a new command-line flag to enable sequencer mode for the Monomer node.
	- Enhanced testing capabilities by adding functionality for setting up and running a verifier node alongside the sequencer node.

### DIFF
--- a/integrations/integrations.go
+++ b/integrations/integrations.go
@@ -54,7 +54,7 @@ import (
 
 const (
 	flagEngineURL         = "monomer.engine-url"
-	flagSequencerMode     = "monomer.sequencer-mode"
+	flagSequencerMode     = "monomer.sequencer"
 	flagDev               = "monomer.dev-start"
 	flagL1AllocsPath      = "monomer.dev.l1-allocs"
 	flagL1DeploymentsPath = "monomer.dev.l1-deployments"
--- a/integrations/integrations.go
+++ b/integrations/integrations.go
@@ -54,6 +54,7 @@ import (
 
 const (
 	flagEngineURL         = "monomer.engine-url"
+	flagSequencerMode     = "monomer.sequencer-mode"
 	flagDev               = "monomer.dev-start"
 	flagL1AllocsPath      = "monomer.dev.l1-allocs"
 	flagL1DeploymentsPath = "monomer.dev.l1-deployments"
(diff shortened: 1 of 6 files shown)


integrations/integrations.go
```diff
@@ -54,6 +54,7 @@ import (
 
 const (
 	flagEngineURL         = "monomer.engine-url"
+	flagSequencerMode     = "monomer.sequencer-mode"
```
joshklop, line 57:
```suggestion
	flagSequencerMode     = "monomer.sequencer"
```
nit

opdevnet/opdevnet_test.go
```diff
@@ -135,6 +135,60 @@ func TestOPDevnet(t *testing.T) {
...
+			break
+		}
+		if i == 29 {
+			t.Fatalf("Verifier failed to sync to block 10. Current block: %v", verifierBlockNum)
+		}
+		time.Sleep(time.Second)
+	}
+
+	// Assert that the verifier and sequencer state roots at block 10 are equal
+	require.Equal(t, verifierL2Backend.BlockChain().GetHeaderByNumber(10).Root, l2Backend.BlockChain().GetHeaderByNumber(10).Root)
```
joshklop, line 191:
```suggestion
	// Wait for the verifier node to sync to block 10
	for i := 0; i < 30; i++ {
		if verifierL2Backend.BlockChain().CurrentHeader().Number >= 10 {
				// Assert that the verifier and sequencer state roots at block 10 are equal
	                        require.Equal(t, verifierL2Backend.BlockChain().GetHeaderByNumber(10).Root, l2Backend.BlockChain().GetHeaderByNumber(10).Root)
		}
		time.Sleep(l1Config.Genesis.BlockTime)
	}
	require.Failf(t, "verifier only synced to block %d", verifierL2Backend.BlockChain().CurrentHeader().Number)
```

joshklop (approved):

opdevnet/opdevnet_test.go
```diff
@@ -135,6 +135,60 @@ func TestOPDevnet(t *testing.T) {
...
+		return nil
+	})
+	require.NoError(t, err)
+	require.NoError(t, verifierL2Node.Start())
+	defer func() {
+		require.NoError(t, verifierL2Node.Close())
+	}()
+	l2EngineURL.IsReachable(ctx)
+
+	verifierOpConfig, err := opdevnet.BuildOPConfig(
```
joshklop, line 161: Any thoughts on putting this in a separate `TestVerifier` test function? It will run in parallel in that case, and might be a bit easier to read.

natebeauregard: I like the idea of separating them out into subtests. I'll look into splitting these up and running them in parallel

opdevnet/opdevnet_test.go
natebeauregard: Unfortunately this will cause the test to always fail since the `require.Failf` case will always be hit. I'll try to think of other ways to handle this case to make the test code more readable

opdevnet/opdevnet_test.go
natebeauregard: I looked into this for a little bit and had a setup that used a channel to store the sequencer state root at block 10 from the existing test for comparison in the new verifier test. Unfortunately there was still a lot of setup logic that had to be copied over to the `TestVerifier` test and I feel like it made it less readable. I'm going to opt to leave this in one test unless you feel strongly about splitting it up

